PR #4, opened 2016-02-01
fix default cacheKey - fixes #3

### LINKED ISSUE #3: issue in default `cacheKey` function
These test fails

```js
test('memoize with multiple primitive arguments', t => {
	let i = 0;
	const memoized = fn(() => i++);
	t.is(memoized('foo'), 0);
	t.is(memoized('foo'), 0);
	t.is(memoized('foo', 'bar'), 1);
	t.is(memoized('foo', 'bar'), 1);
});
```

It always takes `foo` as cache key, even when 2 values are provided. Doing a PR as we speak.

SamVerschueren:
As described in #3, when the first argument was a primitive, it always took that argument as the only key in the cache because of `typeof x !== 'function' && typeof x !== 'object'`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@ module.exports = function (fn, opts) {
 	opts = opts || {};
 
 	var cacheKey = opts.cacheKey || function (x) {
-		if (arguments.length === 1 && x === null || x === undefined || (typeof x !== 'function' && typeof x !== 'object')) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
:sparkles: :)

SamVerschueren:
Woop! Thanks for merging :)

PR #5, opened 2016-07-20
Add stats

Kikobeats:
I want to know your opinion about that.

I added a little stats objects useful under development to know if the memoized is working as you expected.

For avoid too much prints, I added the flag `maxStatsCalls` to setup the number of calls to print the debug message.

Also I added a name per each function to avoid be anonymous and be optimizable for v8.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,15 @@
 'use strict';
-module.exports = function (fn, opts) {
+
+var debug = require('debug')('mem');
+
+function getPercent(partial, total) {
+	return (partial * 100 / total).toFixed(2);
+}
+
+module.exports = function mem(fn, opts) {
 	opts = opts || {};
 
-	var cacheKey = opts.cacheKey || function (x) {
+	var cacheKey = opts.cacheKey || function cacheKey(x) {
 		if (arguments.length === 1 && (x === null || x === undefined || (typeof x !== 'function' && typeof x !== 'object'))) {
 			return x;
 		}
@@ -10,7 +17,17 @@ module.exports = function (fn, opts) {
 		return JSON.stringify(arguments);
 	};
 
-	var memoized = function () {
+	var stats = {
+		hits: 0,
+		miss: 0,
+		total: 0
+	};
+
+	var statsCalls = 0;
+
+	var memoized = function memoized() {
+		++stats.total;
+
 		var cache = memoized.__cache__;
 		var key = cacheKey.apply(null, arguments);
 
@@ -18,9 +35,13 @@ module.exports = function (fn, opts) {
 			var c = cache.get(key);
 
 			if (typeof opts.maxAge !== 'number' || Date.now() < c.maxAge) {
+				++stats.hits;
+				memoized.stats();
 				return c.data;
 			}
 		}
+
+		++stats.miss;
 
 		var ret = fn.apply(null, arguments);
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Why and for what do you need it? What problem does it solve? :)

Kikobeats:
I'm using mem as cache level for avoid cpu cycles in a [string maching lexer/parser](https://github.com/Kikobeats/voll/blob/master/lib/evaluator.js#L3) and I needed to be sure that the rate of hits is good to add mem as dependency.

I choosed mem between other solutions because it haven't dependencies, just a smart wrapper around Map; I modified the core for control my case, but I thought in create this PR because I suppose that know cache stats is a generic and expected behavior that a memoized library could offer.

SamVerschueren:
Actually you could create a more clever map instead and provide it as `cache` option.

```js
class IntelligentMap extends Map {

    constructor(iterable) {
        super(iterable);
        this.stats = {
            hit: 0,
            miss: 0
        }
    }

    get(key) {
        const ret = super.get(key);

        if (ret) {
            this.stats.hit++;
        } else {
            this.stats.miss++;
        }

        return ret;
    }
}
```

Use that as a map and you can achieve the same things as you want with this PR. And you can easily swap it out if you don't need it anymore.

SamVerschueren:
I created something for you [stats-map](https://github.com/SamVerschueren/stats-map). Feel free to suggest extra features through PRs.

Kikobeats:
I like that! more clean. but what is the point? use it at mem? close this PR? say to me guys 😄

sindresorhus:
Good solution. @SamVerschueren Can you add a tip about `stats-map` to the readme?

SamVerschueren:
You can use it with `mem` with the `cache` option.

```js
const mem = require('mem');
const StatsMap = require('stats-map');

const cache = new StatsMap();

const fnMem = mem(fn, {cache});

// Call fnMem()

console.log(cache.stats);
//=> {hits: 10, misses: 7}
```

> Can you add a tip about stats-map to the readme?

👍

SamVerschueren:
Currently, `stats-map` in combination with `mem` will never increment the `misses`. The reason is because I do not extend the `has` method in `stats-map` and do not increment the stats when `has` is called but returns `false`.

I can override the `has` method like this

```js
has(key) {
    const ret = super.has(key);

    if (ret) {
        this.stats.hits++;
    } else {
        this.stats.misses++;
    }

    return ret;
}
```

But that means that in case of `mem`, if it has a `hit`, it will increment the hits by 2 because it will first call `has` (hits+1) and after that it will call `get` (hits+1).

Or should we only increment `misses` if ret is `false`?

```js
has(key) {
    const ret = super.has(key);

    if (!ret) {
        this.stats.misses++;
    }

    return ret;
}
```

sindresorhus:
Hmm, we should actually only count hits in https://github.com/sindresorhus/mem/blob/65d4ab7fb3b43b71b3c968eff65329c8ebb3eee8/index.js#L21, which I don't see how would be possible using your Map subclass. Maybe it's would just be easier to add it here after all.

SamVerschueren:
Actually, I could make that work. If I drop all the logic currently implemented and override the set method like this:

```js
set(key, value) {
    const data = value.data;

    Object.defineProperty(value, 'data', {
        get: () => {
            hits++;
            return data;
        }
    });

    super.set (key, value);
}

get(key) {
    misses++;
    return super.get (key)
}

has (key) {
    const has = super.has(key);

    if (!has) misses++;

    return has;
}
```

This should solve the mem use case. Happy to implement this is stats map.

On mobile, so might have some typos.

sindresorhus:
:+1:

SamVerschueren:
Just released `stats-map@1.0.0` which fixed all of this. I don't think someone will ever use it separately from `mem` because it's coupled to the implementation of `mem`. But I like that this issue got resolved quite elegantly :).

PR #6, opened 2016-07-25
add `stats-map`

SamVerschueren:
@sindresorhus something like this? Or do you want to place it somewhere else?


sindresorhus:
I was thinking a `Tip` section with a link and a example.

SamVerschueren:
Fixed it. But let's first discuss the issue I have currently in #5 before merge.

PR #7, opened 2017-03-03
Add support for clearing cache for specific function arguments

quocnguyen:
Sometime due to bad connection `got` return the rejected promise (timeout), which is not good to be cached. 
It would be nice to be able to clear ONLY that data while the others still remain untouched.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,7 +18,7 @@ module.exports = (fn, opts) => {
 	}, opts);
 
 	const memoized = function () {
-		const cache = cacheStore.get(memoized);
+		const {cache} = cacheStore.get(memoized);
 		const key = opts.cacheKey.apply(null, arguments);
 
 		if (cache.has(key)) {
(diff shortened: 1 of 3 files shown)


sindresorhus:
I think it would be better to just not cache rejected promises as all. Interested in doing a pull request for that instead?

quocnguyen:
It was not only about the rejected promise, imo the use case for this pull request is to bring more control over cached stuff.

for example: A UI where user can find all cached function together with their params, they can select an item in that list and click remove manually.

sindresorhus:
Ok. I can see the use-case for this. Can you remove the destructuring and default parameter syntax usage? I still support Node.js 4, which doesn't have those.

sindresorhus:
> Sometime due to bad connection got return the rejected promise (timeout), which is not good to be cached.

I opened an issue about this here: https://github.com/sindresorhus/mem/issues/8 I still think that's the correct solution for this problem.

PR #10, opened 2017-09-20
Don't cache rejected promises - fixes #8

### LINKED ISSUE #8: Don't cache a rejected promise
I think there should still be an option to cache a reject promise.

SamVerschueren:
This PR fixes #8 by not caching a rejected promise by default. I also added a `cacheRejection` option (couldn't come up with a better name) to re-enable it.

Feedback welcome!

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,7 @@ module.exports = (fn, opts) => {
 	opts = Object.assign({
 		cacheKey: defaultCacheKey,
 		cache: new Map(),
-		cacheRejection: false
+		cachePromiseRejection: false
 	}, opts);
 
 	const memoized = function () {
@@ -33,21 +33,22 @@ module.exports = (fn, opts) => {
 
 		const ret = fn.apply(null, arguments);
 
-		if (isPromise(ret) && opts.cacheRejection === false) {
-			// Only cache resolved promises unless `cacheRejection` is set to `true`
+		const setData = (key, data) => {
+			cache.set(key, {
+				data,
+				maxAge: Date.now() + (opts.maxAge || 0)
+			});
+		};
+
+		if (isPromise(ret) && opts.cachePromiseRejection === false) {
+			// Only cache resolved promises unless `cachePromiseRejection` is set to `true`
 			ret
 				.then(() => {
-					cache.set(key, {
-						data: ret,
-						maxAge: Date.now() + (opts.maxAge || 0)
-					});
+					setData(key, ret);
 				})
 				.catch(() => { });
 		} else {
-			cache.set(key, {
-				data: ret,
-				maxAge: Date.now() + (opts.maxAge || 0)
-			});
+			setData(key, ret);
 		}
 
 		return ret;
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 'use strict';
 const mimicFn = require('mimic-fn');
+const isPromise = require('p-is-promise');
 
 const cacheStore = new WeakMap();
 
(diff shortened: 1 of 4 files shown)


kevva:
I like `cacheRejection` or `cacheRejected`.

index.js
```diff
@@ -31,10 +33,22 @@ module.exports = (fn, opts) => {
...
+			ret
+				.then(() => {
+					cache.set(key, {
+						data: ret,
+						maxAge: Date.now() + (opts.maxAge || 0)
+					});
+				})
+				.catch(() => { });
+		} else {
+			cache.set(key, {
```
sindresorhus: This is duplicated. Can you extract it into an arrow function above the `if` statement on line 36?

readme.md
```diff
@@ -106,6 +106,13 @@ Default: `new Map()`
 
 Use a different cache storage. Must implement the following methods: `.has(key)`, `.get(key)`, `.set(key, value)`, and optionally `.clear()`. You could for example use a `WeakMap` instead or [`quick-lru`](https://github.com/sindresorhus/quick-lru) for a LRU cache.
 
+##### cacheRejection
+
+Type: `boolean`<br>
+Default: `false`
+
+Set to `true` if you want to cache rejected promises.
```
sindresorhus: I think `Set to true if you want to` is moot. The default is defined, so we can just describe what it does:

> Cache rejected promises.

test.js
```diff
@@ -74,6 +74,40 @@ test('promise support', async t => {
...
+test('do not cache rejected promises', async t => {
+	let i = 0;
+	const memoized = m(() => {
+		i++;
+		if (i === 1) {
+			return Promise.reject(new Error('foo bar'));
+		}
+
+		return Promise.resolve(i);
+	});
```
sindresorhus: This can be simplified by using an async function:

```js
const memoized = m(async () => {
	i++;

	if (i === 1) {
		throw new Error('foo bar');
	}

	return i;
}, {
	cacheRejection: true
});
```

sindresorhus:
Let's go with `cachePromiseRejection` as `cacheRejection` can sound like it's about the rejection of the cache, not caching of the rejected promise.

test.js
SamVerschueren: Much better :)

SamVerschueren:
Updated

sindresorhus:
Perfect 👌🌟

LinusU:
Hmm, it _seems_ like this will wait to cache the value until the promise have settled, which is not how this module used to behave. I'm currently using the module something like this:

```js
const getPlanInfo = mem(async (planName) => {
  const input = await expensiveApiCallThatFetchesPlanInfo()

  return { currency: input.currency.toUpperCase(), amount: input.amount }
})
```

In 1.x every request for a specific plan would all return the same Promise and only one request would go out to the remote server. In 2.x there will be one request sent per call until the first Promise settles.

`p-memoize` works by removing the Promise from the cache if the promise rejects (actually, it _seems_ like it clears the entire cache when any promise rejects, since there is no way to clear a specific entry from the cache), which lets all the first requests use the single outgoing request to my third party service.

-----

Would a PR to switch to returning the same pending `Promise` for all subsequent call until the promise settles be accepted?

sindresorhus:
> Would a PR to switch to returning the same pending Promise for all subsequent call until the promise settles be accepted?

Yes. This was just an oversight.

LinusU:
Great 👍 

There is one small problem though 🤔 

There is currently no specification that the provided `cache` must support `.delete`, even `.clear` is optional.

How do you want to solve this, should we add that it should provide `.delete` (which `Map` already does)?

LinusU:
Submitted #12 which requires `.delete` to be present on the `cache` object

LinusU:
ping @sindresorhus 😄

PR #11, opened 2017-09-20
Add support for prototype method functions - fixes #9

### LINKED ISSUE #9: Enable memoization for prototype methods
I managed to use this library with prototype methods by passing `this` instead of `null` in `apply` function at `https://github.com/sindresorhus/mem/blob/master/index.js#L32`.

It allows me to use it like this:

```
function Foo() {
    this.myvalue = 'val';
}
var bar = function () {
    console.log(this.myvalue);
    // Display `val`
};
Foo.prototype.bar = mem(bar)
```
Since I do not know if it is safe, I didn't submit a pull request. Maybe it should be passed through options?

If changing `this` for `null` is ok, I can submit a PR.

SamVerschueren:
Fixes #9.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,7 +29,7 @@ module.exports = (fn, opts) => {
 			}
 		}
 
-		const ret = fn.apply(null, arguments);
+		const ret = fn.apply(this, arguments);
 
 		cache.set(key, {
 			data: ret,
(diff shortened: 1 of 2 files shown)


kevva:
Perfect.

PR #12, opened 2017-09-28
Cache pending promises

LinusU:
Fixes issue discussed in #10 

This should probably go out as a semver major since we are putting additional requirements on the cache object and since the previous behaviour, although not intentionally added, might already be relied on...

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,15 +40,11 @@ module.exports = (fn, opts) => {
 			});
 		};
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
Looks good. Thank you. And sorry about the delay :)

LinusU:
No problem, cheers 🍻

PR #17, opened 2018-02-21
Memoize calls to mem itself

keithamus:
This way memoize can wrap functions across multiple files in a single
realm but still share a cache between those calls. Consider:

```js
// file A:
const foo = require('foo')
const memfoo = mem(foo)
``

```js
// file B:
const foo = require('foo')`
const memfoo = mem(foo)
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,10 @@
 const mimicFn = require('mimic-fn');
 const isPromise = require('p-is-promise');
 
+// Store functions that memoize has previously memoized,
+// so that memoize calls can be cached
+const fnStore = new WeakMap();
+
 const cacheStore = new WeakMap();
 
 const defaultCacheKey = (...args) => {
@@ -25,6 +29,14 @@ module.exports = (fn, options) => {
 		cache: new Map(),
 		cachePromiseRejection: false
 	}, options);
+
+	const fnKey = JSON.stringify(opts);
+	if (fnStore.has(fn)) {
+		const optionsStore = fnStore.get(fn);
+		if (optionsStore.has(fnKey)) {
+			return optionsStore.get(fnKey);
+		}
+	}
 
 	const memoized = function (...args) {
 		const cache = cacheStore.get(memoized);
@@ -62,6 +74,10 @@ module.exports = (fn, options) => {
 	mimicFn(memoized, fn);
 
 	cacheStore.set(memoized, options.cache);
+	if (!fnStore.has(fn)) {
+		fnStore.set(fn, new Map());
+	}
+	fnStore.get(fn)[fnKey] = memoized;
 
 	return memoized;
 };
--- a/test.js
+++ b/test.js
@@ -17,6 +17,62 @@ test('memoize', t => {
 	t.is(memoized('foo', 'bar'), 2);
 });
 
+test('memoize itself is memoized', t => {
+	let i = 0;
+	const f = () => i++;
+	let memoized = m(f);
+	t.is(memoized(), 0);
+	t.is(memoized(), 0);
+	memoized = m(f);
+	t.is(memoized(), 0);
+	t.is(memoized('foo'), 1);
+	t.is(memoized('foo'), 1);
+	memoized = m(f);
+	t.is(memoized('foo'), 1);
+	t.is(memoized('foo', 'bar'), 2);
+	t.is(memoized('foo', 'bar'), 2);
+	memoized = m(f);
+	t.is(memoized('foo', 'bar'), 2);
+
+	// Resolves to same options bag, use memoized version
+	memoized = m(f, {cachePromiseRejection: false});
+	t.is(memoized(), 1);
+	t.is(memoized('foo'), 1);
+	t.is(memoized('foo', 'bar'), 2);
+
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -53,6 +62,10 @@ module.exports = (fn, opts) => {
 	mimicFn(memoized, fn);
 
 	cacheStore.set(memoized, opts.cache);
+	if (!fnStore.has(fn)) {
+		fnStore.set(fn, {});
```
sindresorhus: I think the `{}` would be better as a `new Map()` instead.

index.js
```diff
@@ -19,6 +20,14 @@ module.exports = (fn, opts) => {
 		cachePromiseRejection: false
 	}, opts);
 
+	const fnKey = JSON.stringify(opts);
```
sindresorhus: `JSON.stringify` cannot handle `new Map()`, so if `options.cache` differs between calls, it will still use the same instance.

```js
JSON.stringify({foo: new Map([['a', 'b']])})
//=> '{"foo":{}}'
```

index.js
```diff
@@ -19,6 +20,14 @@ module.exports = (fn, opts) => {
...
+	if (fnStore.has(fn)) {
+		const optStore = fnStore.get(fn);
```
sindresorhus: `optStore` => `optionsStore`

index.js
```diff
@@ -2,6 +2,7 @@
 const mimicFn = require('mimic-fn');
 const isPromise = require('p-is-promise');
 
+const fnStore = new WeakMap();
```
sindresorhus: I think this could use a succinct comment about its purpose, so it's clearer when you read the code that it's for caching calls to mem itself.

test.js
```diff
@@ -17,6 +17,24 @@ test('memoize', t => {
...
+	t.is(memoized(), 0);
+	t.is(memoized('foo'), 1);
+	t.is(memoized('foo'), 1);
+	memoized = m(f);
+	t.is(memoized('foo'), 1);
+	t.is(memoized('foo', 'bar'), 2);
+	t.is(memoized('foo', 'bar'), 2);
+	memoized = m(f);
+	t.is(memoized('foo', 'bar'), 2);
+});
```
sindresorhus: Needs some more tests using different options, especially testing with the `cache` option.

sindresorhus:
@LinusU Thoughts?

index.js
keithamus: Agreed. Ignoring the `cache` option is _kind of intentional_ though - the intent is to keep `mem(fn)` idempotent; giving `mem(fn)` its own cache each time is redundant. It could even be said giving `mem(fn)` and `mem(fn, { cachePromiseRejection: false })` their own cache is also redundant.

There's also another issue we have; `options = Object.assign(...)` will always generate a new reference, and thus just putting `options` in a Map will mean `mem` will never memoize. So we need some kind of mechanism to cache these options, but yes - as you say - JSON.stringify is insufficient.

One way around this could be to store each option in a Map - this is a little clunky, but it also gives us the benefit of ignoring the `cache` as part of the memoizing.

```js
	if (fnStore.has(fn)) {
		// Check if a previously memoized `fn` was given the same cacheKey option:
		const cacheKeyStore = fnStore.get(fn);
		if (cacheKeyStore.has(options.cacheKey)) {
			// Check if a previously memoized `fn` was given the same cachePromiseRejection option:
			const cachePromiseRejectionStore = cacheKeyStore.get(options.cacheKey)
			if (cachePromiseRejectionStore.has(options.cachePromiseRejection)) {
				// We have a memoized version stored with all the same options, return it
				return cachePromiseRejectionStore.get(options.cachePromiseRejection)
			}
		}
	}
```

keithamus:
I've addressed most comments, but the key is still based on `JSON.stringify`. Not sure of how to address that in a reasonable way. Feedback welcome on that.

index.js
```diff
@@ -26,6 +30,14 @@ module.exports = (fn, options) => {
 		cachePromiseRejection: false
 	}, options);
 
+	const fnKey = JSON.stringify(opts);
```
LinusU: `JSON.stringify(options)`?

index.js
```diff
@@ -62,6 +74,10 @@ module.exports = (fn, options) => {
 	mimicFn(memoized, fn);
 
 	cacheStore.set(memoized, options.cache);
+	if (!fnStore.has(fn)) {
+		fnStore.set(fn, new Map());
+	}
+	fnStore.get(fn)[fnKey] = memoized;
```
LinusU: `fnStore.get(fn).set(fnKey, memoized);`?

keithamus:
> Maybe a good start would be to only cache calls to mem without an options object. I think that that is the most common use case in any way, although I have no data to back that up 🙈

This is definitely my use-case. I'd be happy doing this if it would be considered for merging.

sindresorhus:
Closing this for lack of activity, but I'm still happy to take a PR for:

> Maybe a good start would be to only cache calls to mem without an options object. I think that that is the most common use case in any way, although I have no data to back that up 🙈

But consideration should be given about possible memory leaks.

PR #19, opened 2018-08-22
Automatically release memory when an item expires - fixes #14

### LINKED ISSUE #14: Memory leak: old results are not deleted from the cache

SamVerschueren:
This PR fixes #14 by using [map-age-cleaner](https://github.com/SamVerschueren/map-age-cleaner) which removes item when it expires.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict';
 const mimicFn = require('mimic-fn');
 const isPromise = require('p-is-promise');
+const mapAgeCleaner = require('map-age-cleaner');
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
Can you fix the merge conflict?

SamVerschueren:
Done

mbrookes:
Any possibility if backporting this to 3.0? https://github.com/yargs/yargs/issues/1223 Thanks!

PR #24, opened 2018-11-14
Unref timeouts to let process exit properly

johansteffner:


test.js
```diff
@@ -111,22 +111,22 @@ test('cache option', t => {
 
 test('promise support', async t => {
 	let i = 0;
-	const memoized = m(async () => i++);
+	const memoized = m(() => Promise.resolve(i++));
```
SamVerschueren: Why did you change this test file?

package.json
```diff
@@ -32,7 +32,7 @@
 		"promise"
 	],
 	"dependencies": {
-		"map-age-cleaner": "^0.1.1",
+		"map-age-cleaner": "^0.1.3",
```
SamVerschueren: It should automatically be updated when reinstalling `mem` because of semver.

johansteffner: Yes, but when using tools like [renovate](https://renovatebot.com/), changes in a packages dependencies won't be detected.

test.js
johansteffner: The xo linter complained about having an async function without an await inside. It's specified as `"xo": "*"` in dependencies, so i guess there was some change of this rule in some newer version.

package.json
LinusU: @SamVerschueren So, when it actually fixes a bug, I think it's appropriate to release a new version with the minimum version bumped (with the `^` specifier) since the package isn't actually compatible with `0.1.1` and `0.1.2` in the sense that they are affected by the bug and thus are behaving buggy when consumed by someone else. Therefore I think that this should be merged and released as a patch release.

@johansteffner in the meantime you can bump just that package in your lock-file by running `npm update map-age-cleaner --depth=50` to just bump that package across your entire lock-file.

<details><summary>off topic:</summary>

I actually think that the only way to properly do automatic updates with Renovate and similar tools is to bump _the lock-file entry only_, whenever _any_ dependency is updated. There will be _a ton_ of bumps, but it's the only way to know exactly which dependency that broke your build.

If you only bump in you `package.json` file and let the package-lock regenerate itself you will consolidate many updated packages into only one PR + build; and you have only arbitrarily picked which packages you'll bump on...

</details>

johansteffner: @SamVerschueren do you agree with @LinusU reasoning about releasing this? Otherwise I'll close this PR.

PR #27, opened 2019-02-23
Avoid `Date.now()` call when `maxAge` is not enabled

SoyYoRafa:
Avoids relatively expensive Date.now call when maxAge is not enabled. Only enable mapAgeCleaner when options.maxAge is a positive non-zero number.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@ module.exports = (fn, options) => {
 		cachePromiseRejection: false
 	}, options);
 
-	if (typeof options.maxAge === 'number') {
+	if (typeof options.maxAge === 'number' && options.maxAge > 0) {
 		mapAgeCleaner(options.cache);
 	}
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -31,17 +31,18 @@ module.exports = (fn, options) => {
 		cachePromiseRejection: false
 	}, options);
 
-	if (typeof options.maxAge === 'number') {
+	if (typeof options.maxAge === 'number' && options.maxAge > 0) {
```
sindresorhus: If the user specifies `{maxAge: 0}` we have a memory leak as it will never clean up those entries.

rusty0412: I assumed the intended behavior of `{maxAge: 0}` would be to disable age-based eviction. Why would someone memoize with `{maxAge: 0}`? I'm happy to change the condition to `options.maxAge >= 0`

SoyYoRafa: I assumed the intended behavior of `{maxAge: 0}` would be to disable age-based eviction. Why would someone memoize with `{maxAge: 0}`? I'm happy to change the condition to `options.maxAge >= 0`

sindresorhus: > I assumed the intended behavior of {maxAge: 0} would be to disable age-based eviction.

Assumptions are dangerous. Someone could also assume it causes it to be evicted right away.

> Why would someone memoize with {maxAge: 0}?

I dunno, but from experience, users do a lot of things you don't assume they would do.

>  I'm happy to change the condition to options.maxAge >= 0

👍 Also document what setting `0` does and I think we should also document and validate that the number is non-negative, to prevent mistakes.

SoyYoRafa: Done

PR #29, opened 2019-03-14
TypeScript - Fix the cache type

dylang:
@sindresorhus Thanks for adding types to your repos! 

I believe the `cache` type needs to include the `data`, otherwise this isn't possible:

```typescript

const cache = new QuickLRU<string, { data: Bundle }>({
    maxSize: BUNDLE_COUNT_TO_MEMORIZE
});

const loadJsonFileMemorized = mem(
    (filename: string) => loadJsonFile<Bundle>(filename),
    {
        cache
    }
);

// After modifying we want to update the cache directly to avoid another load.
cache.set(filename, { data: updatedBundle });
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -33,7 +33,7 @@ export interface Options<
 	 *
 	 * @default new Map()
 	 */
-	readonly cache?: CacheStorage<CacheKeyType, { data: ReturnType; maxAge?: number; }>;
+	readonly cache?: CacheStorage<CacheKeyType, { data: ReturnType; maxAge: number; }>;
 
 	/**
 	 * Cache rejected promises.
--- a/index.d.ts
+++ b/index.d.ts
@@ -33,7 +33,7 @@ export interface Options<
 	 *
 	 * @default new Map()
 	 */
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -33,7 +33,7 @@ export interface Options<
 	 *
 	 * @default new Map()
 	 */
-	readonly cache?: CacheStorage<CacheKeyType, ReturnType>;
+	readonly cache?: CacheStorage<CacheKeyType, { data: ReturnType; maxAge?: number; }>;
```
BendingBender: Why is `maxAge` optional? It is not being set conditionally in the code.

dylang: If you aren't using the `maxAge` option then this value is not necessary.

BendingBender: With this definition, you're basically saying that `maxAge` _may_ be sometimes provided to the underlying API. This is, however, not the case. `maxAge` is _always_ set, disregarding of whether you use it or not.

When you implement a custom cache storage you don't actually have to use the `maxAge` value, it is however, always there.

dylang: Okay, thanks for that explanation @BendingBender! Changing `maxAge` to required.

PR #30, opened 2019-03-15
`mem` is supposed to be callable when `require`'d, in addition to being a `default`

weswigham:
Without this the definition file breaks JS intellisense in projects like `prettier`.

@sindresorhus do you have any other cjs packages that do both an `module.exports` assignment and a default member? They should all have definitions that work similarly to this.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,46 +1,48 @@
-export interface CacheStorage<
-	KeyType extends unknown,
-	ValueType extends unknown
-> {
-	has(key: KeyType): boolean;
-	get(key: KeyType): ValueType | undefined;
-	set(key: KeyType, value: ValueType): void;
-	delete(key: KeyType): void;
-	clear?: () => void;
-}
+declare namespace mem {
+	export interface CacheStorage<
+		KeyType extends unknown,
+		ValueType extends unknown
+	> {
+		has(key: KeyType): boolean;
+		get(key: KeyType): ValueType | undefined;
+		set(key: KeyType, value: ValueType): void;
+		delete(key: KeyType): void;
+		clear?: () => void;
+	}
 
-export interface Options<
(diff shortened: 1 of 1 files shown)


weswigham:
For reference, we found this because the last update of `mem` triggered some new errors in our [user code test suite](https://github.com/Microsoft/TypeScript/pull/30427/files#diff-451edb7ad29bc4adf4241b99df19b5e7R13).

sindresorhus:
The TS definition is intentional and what I use on all my other projects.

---

> Without this the definition file breaks JS intellisense in projects like prettier.

You have to import it using `const mem = require('mem').default;` if you want JS intellisense. I support `const mem = require('mem');` for convenience of CommonJS users, but VS Code seems to not pick it up. I've decided that it's worth it to have proper ES2015 module typings.

weswigham:
> but VS Code seems to not pick it up. 

VS Code doesn't pick it up _because it uses this file_.

weswigham:
> I've decided that it's worth it to have typings with the least amount of TS-specific syntax

That's a mistake. The extra syntax exists in declaration files expressly for expressing the shapes commonjs modules can produce but esm cannot. So long as the source isn't esm or TS, you should be using whatever syntax allows the .d.ts shape to match the runtime shape.

This isn't an opinion - I'm on the TS team. `.d.ts` files are important for type information across both JS and TS, and their accuracy is paramount. People lose trust in their editor if its completion/error behavior doesn't match reality. If you don't want to maintain an accurate `.d.ts`, you shouldn't be bundling it with your project and you should let `DefinitelyTyped` take care of it.

weswigham:
If it's a matter of legitimately considering the callable root namespace as deprecated (for all consumers), I can recommend placing an `@deprecated` jsdoc on the signature and only removing it once it's removed from the accompanying source.

PR #35, opened 2019-05-09
Cache function reference when it's a single argument

bfred-it:
Since the very first version, mem has a _"cache if it's not a function AND not an object"_, which I don't understand because functions are can't be serialized via `JSON.stringify`

In short (unless I'm drunk 🤔) I think `mem` should _always_ use the ONE argument as key, unless it's a serializable object.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,9 @@ const defaultCacheKey = (...arguments_) => {
 
 	if (arguments_.length === 1) {
 		const [firstArgument] = arguments_;
-		if (typeof firstArgument !== 'object' || firstArgument === null) {
+		const isObject = typeof firstArgument === 'object' && firstArgument !== null;
+		const isPrimitive = !isObject;
+		if (isPrimitive) {
 			return firstArgument;
 		}
 	}
--- a/index.js
+++ b/index.js
@@ -15,7 +15,7 @@ const defaultCacheKey = (...arguments_) => {
 		if (
 			firstArgument === null ||
 			firstArgument === undefined ||
-			typeof firstArgument === 'function'
+			typeof firstArgument !== 'object'
 		) {
 			return firstArgument;
 		}
--- a/index.js
+++ b/index.js
@@ -15,7 +15,7 @@ const defaultCacheKey = (...arguments_) => {
 		if (
 			firstArgument === null ||
 			firstArgument === undefined ||
-			(typeof firstArgument !== 'function' && typeof firstArgument !== 'object')
+			typeof firstArgument === 'function'
 		) {
 			return firstArgument;
 		}
--- a/index.js
+++ b/index.js
@@ -12,11 +12,7 @@ const defaultCacheKey = (...arguments_) => {
 
 	if (arguments_.length === 1) {
 		const [firstArgument] = arguments_;
-		if (
-			firstArgument === null ||
-			firstArgument === undefined ||
-			typeof firstArgument !== 'object'
-		) {
+		if (firstArgument === null || typeof firstArgument !== 'object') {
 			return firstArgument;
 		}
 	}
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ const defaultCacheKey = (...arguments_) => {
 
 	if (arguments_.length === 1) {
 		const [firstArgument] = arguments_;
-		if (firstArgument === null || typeof firstArgument !== 'object') {
+		if (typeof firstArgument !== 'object' || firstArgument === null) {
 			return firstArgument;
 		}
 	}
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -12,11 +12,7 @@ const defaultCacheKey = (...arguments_) => {
 
 	if (arguments_.length === 1) {
 		const [firstArgument] = arguments_;
-		if (
-			firstArgument === null ||
-			firstArgument === undefined ||
-			(typeof firstArgument !== 'function' && typeof firstArgument !== 'object')
-		) {
+		if (typeof firstArgument !== 'object' || firstArgument === null) {
```
sindresorhus: It took me a while to realize what this code does. think this would make it more readable:

```suggestion
		const isObject = typeof firstArgument === 'object' && firstArgument !== null;
		const isPrimitive = !isObject;
		if (isPrimitive) {
```

sindresorhus:
I think the reason I didn't support functions is that the user can set their own `cache`, and it might be something that needs to serialize the value.

bfred-it:
> and it might be something that needs to serialize the value.

In that case they can also set `cacheKey`, since it goes hand in hand with `cache`'s key-ing mechanism.

index.js
bfred-it: Now I'm not sure that we need the extra `isPrimitive` step. `!isObject` makes enough sense on its own.

sindresorhus:
> In that case they can also set cacheKey, since it goes hand in hand with cache's key-ing mechanism.

I agree. Can you add a note about that in the readme? Just so we're clear about how it works. It might not be obvious to everyone that we cache the function reference too.

bfred-it:
> It might not be obvious to everyone that we cache the function reference too.

Why would it be otherwise? I think at most we could mention that _objects and multiple arguments_ are cached by value.

index.js
sindresorhus: I disagree. It's not obvious to everyone that `!isObject` means it's a primitive. Doesn't hurt to make it perfectly clear.

readme.md
```diff
@@ -2,7 +2,7 @@
 
 > [Memoize](https://en.wikipedia.org/wiki/Memoization) functions - An optimization used to speed up consecutive function calls by caching the result of calls with identical input
 
-Memory is automatically released when an item expires.
+The input is cached by reference when it’s not an object and when the argument is only one; it’s cached by value in all other cases (where `JSON.stringify` supports it). In the first case, memory is automatically released when an item expires (thanks to `WeakMap`).
```
sindresorhus: > The input is cached by reference when it’s not an object

This can never be true, as "not object" means a primitive, which cannot be a reference.

bfred-it: lol true

bfred-it: I see what you meant earlier now

index.js
bfred-it: I see that `primitive` is also mentioned in the readme, so it sorta makes sense to clarify it here as well.

readme.md
```diff
@@ -2,7 +2,7 @@
 
 > [Memoize](https://en.wikipedia.org/wiki/Memoization) functions - An optimization used to speed up consecutive function calls by caching the result of calls with identical input
 
-Memory is automatically released when an item expires.
+The input is cached by value (if supported by `JSON.stringify`) unless it’s a single argument of type `function`, in which case it’s stored by reference. Memory is automatically released when an item expires.
```
sindresorhus: I think it should be documented here instead: https://github.com/sindresorhus/mem#cachekey

bfred-it: Indeed, I noticed that there's plenty of overlap though, to the point that the comment about functions is the only missing part, and even then it's implied by "used directly as key"

readme.md
```diff
@@ -101,7 +101,7 @@ Milliseconds until the cache expires.
 
 Type: `Function`
 
-Determines the cache key for storing the result based on the function arguments. By default, if there's only one argument and it's a [primitive](https://developer.mozilla.org/en-US/docs/Glossary/Primitive), it's used directly as a key, otherwise it's all the function arguments JSON stringified as an array.
+Determines the cache key for storing the result based on the function arguments. By default, if there's only one argument and it's a [primitive](https://developer.mozilla.org/en-US/docs/Glossary/Primitive), it's used directly as a key (if it's a `function`, its reference will be used as key), otherwise it's all the function arguments JSON stringified as an array.
```
sindresorhus: https://github.com/sindresorhus/mem/blob/1b306a2c2ea1f3f72e1453a9f69f17ae2e9c26de/index.d.ts#L22-L26 too

bfred-it:
Will you release this as a patch or are you waiting for #36 to release a major?

PR #37, opened 2019-05-11
Drop `maxAge`

bfred-it:
Similarly to #36, `maxAge` seems just _one_ of the ways to keep the cache clean and its implementation affects _all_ users.

I suggest dropping the option and letting the user provide an [expiry-map](https://github.com/SamVerschueren/expiry-map) (by the same author of `max-age-cleaner`) or [quickl-lru](https://github.com/sindresorhus/quick-lru) if needed, which is pretty straight forward thanks to the `cache` option.

Also closes #27


sindresorhus:
It's provided for convenience as it's a common use-case. I don't really see the benefit of dropping it. 

> maxAge seems just one of the ways to keep the cache clean and its implementation affects all users.

Does it affect them in a negative way though?

I'm struggling to understand the motivation of this PR.

bfred-it:
> I'm struggling to understand the motivation of this PR.

- 
- the memory saving of not having to wrap all values in an object
- module complexity
- adding the extra module is almost as easy as as enabling this feature, if needed at all

sindresorhus:
> the memory saving of not having to wrap all values in an object

I don't think this matters in real-world scenarios at all.

> module complexity

It's like 5 lines in total.

> adding the extra module is almost as easy as as enabling this feature, if needed at all

You then remove the "module complexity" and instead move it to be "docs complexity" and "user complexity".

bfred-it:
How are the docs more complex? It just needs a `and expiry-map` next to the lru map mention. 

It’s simply a “it doesn’t belong here” situation, just like quick-lru isn’t part of this module. Either way it’s not a big deal 🙂

readme.md
```diff
@@ -55,26 +53,6 @@ const memoized = mem(counter);
 })();
 ```
 
-```js
-const mem = require('mem');
-const got = require('got');
-const delay = require('delay');
-
-const memGot = mem(got, {maxAge: 1000});
```
bfred-it: Or I could restore this example with

```diff
+  const ExpiryMap = require('expiry-map');

+  const memGot = mem(got, {cache: new ExpiryMap(1000)});
-  const memGot = mem(got, {maxAge: 1000});
```

sindresorhus:
It also complicates it if you want a custom cache that is not a Map. Then you need to manually implement `maxAge` yourself, or create a wrapper for your chosen data structure. I just feel that it's something many users would need, especially when using `mem` with HTTP requests, that it's worth having it built-in.

bfred-it:
> It also complicates it if you want a custom cache that is not a Map

That's the map implementations' fault. Outside of this module, I currently can't combine `quick-lru` and `expiry-map` for example.


> I just feel that it's something many users would need

That's the reason then

sindresorhus:
> Outside of this module, I currently can't combine quick-lru and expiry-map for example.

I'm happy to support passing in a custom map to `quick-lru`. Open an issue.

PR #39, opened 2019-05-21
Cleanup

bfred-it:
- avoid repeating `options.`
- inline `setData` function
- avoid `mapAgeCleaner` wrapper, closes #27

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,7 +49,6 @@ const mem = (fn, options = {}) => {
 		} : cacheItem);
 
 		if (isPromise(cacheItem) && cachePromiseRejection === false) {
-			// Remove rejected promises from cache if `cachePromiseRejection` is set to `false`
 			cacheItem.catch(() => cache.delete(key));
 		}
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -22,41 +22,34 @@ const defaultCacheKey = (...arguments_) => {
...
-
-	if (typeof options.maxAge === 'number') {
-		mapAgeCleaner(options.cache);
+const mem = (fn, options = {}) => {
+	const {
+		cacheKey = defaultCacheKey,
+		cache = new Map(),
+		cachePromiseRejection = true,
+		maxAge
+	} = options;
```
bfred-it: `options` can be completely avoided too:

```js
const mem = (fn, {
	cacheKey = defaultCacheKey,
	cache = new Map(),
	cachePromiseRejection = true,
	maxAge
} = {}) => {
	...
```

bfred-it: `options` can be completely avoided too:

```js
const mem = (fn, {
	cacheKey = defaultCacheKey,
	cache = new Map(),
	cachePromiseRejection = true,
	maxAge
} = {}) => {
	...
```

PR #43, opened 2019-07-24
Simplify default serializer; drop promise-related option

### LINKED ISSUE #20: Get wrong results when use regexp or Symbol as function arguments

### LINKED ISSUE #40: Expectations, ease of use and clarity
# Demo

1. Object will be memoized:

```js
run({str: 'hello'})
```

2. Object will be memoized, but not really: #20 

```js
run({str: 'hello', reg: /a/})
```

3. Object will be memoized, but not at all like one would expect:

```js
run(document.querySelector('a'))
// Hint: Element is an object, so it will be cached as `"{}"`
// which is the result of `JSON.stringify(element)` 
```

4. Functions will be memoized:

```js
run(copy)
```

5. Functions will be memoized, until they're not:

```js
run(copy, options)
```

# Solutions

1. Fine

2. `{cacheKey: serializeJavascript}`, a more comprehensive object serializer, as suggested in https://github.com/sindresorhus/mem/issues/38

3. `{cacheKey: a => a}`, which means "store the first argument as reference"

4. Fine

5. `{cache: new ManyKeysMap(), cacheKey: (...args) => args}` using [`many-keys-map`](https://github.com/bfred-it/many-keys-map) (as previously suggested in https://github.com/sindresorhus/mem/pull/33/files)

There's no single way to do things, but the current situation tries to fit specific cases (e.g. `run(optionsObject)`) that don't work well in all situations even if they _appear_ to work (because **the cache logic isn't very simple**)

So perhaps, instead of trying to be smart and understand the type of object the function received, it should default to `cacheKey: a => a` and offer stringification regardless of argument count via:

- a new `cacheByValue: true` option
- suggesting 
  - `cacheKey: JSON.stringify`, which is very clear an lightweight, and
  - `cacheKey: serializeJavascript` as a more comprehensive option, either as a dependency (via `cacheKey: mem.serialize`) or external (via `cacheKey: require('serialize-javascript')`)

fregante:
Implements and closes #40 
Replaces and closes #22
Fixes #20
(Basically merging this PR will close everything in this repo 🎉)

In short, what used to be known as `cacheKey: x => x` is now the default.

This default aligns `mem` with other common memoize implementations (like lodash) while still supporting and suggesting a variety of serialization methods which the user can pick.

This makes the memoization logic very straightforward, explicit and fast by default. Refer to [the new readme](https://github.com/sindresorhus/mem/tree/ede5403a634cbe40fc8e36b71644ceebf439501e#cachekey)

---

The main API change is the new signature of `cacheKey`, which instead of receiving the arguments as `cacheKey(...arguments_)`, it receives them as an array. This change makes it easy to supply a plain serializer like:

```js
mem(() => i++, {cacheKey: JSON.stringify});
mem(() => i++, {cacheKey: serializeJavascript});
```

Instead of having to wrap & lengthen the two most likely implementations:

```js
mem(() => i++, {cacheKey: (...args) => JSON.stringify(args)}); // 😬
mem(() => i++, {cacheKey: (...args) => serializeJavascript(args)}); // 😖
```

---

Since this will be a breaking change, I dropped `cachePromiseRejection` since it makes more sense as part of `p-memoize`. This change also makes it easy to copy `mem` into `p-memoize` [as suggested by Sindre](https://github.com/sindresorhus/p-memoize/issues/3#issuecomment-405025314) instead of having to maintain an increasingly-complex in-house stringifier [like what's in #22](https://github.com/sindresorhus/mem/blob/4d72500730c9cb40f8b40e38e1f4c8bbdafba18e/index.js#L6-L43)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,16 +20,34 @@ declare namespace mem {
 		readonly maxAge?: number;
 
 		/**
-		Determines the cache key for storing the result based on the function arguments. By default, if there's only one argument and it's a [primitive](https://developer.mozilla.org/en-US/docs/Glossary/Primitive), it's used directly as a key (if it's a `function`, its reference will be used as key), otherwise it's all the function arguments JSON stringified as an array.
+		Determines the cache key for storing the result based on the function arguments. By default, **only the first argument is considered** and it only works with [primitives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive).
 
-		You could for example change it to only cache on the first argument `x => JSON.stringify(x)`.
+		A `cacheKey` function can return any type supported by `Map` (or whatever structure you use in the `cache` option).
+
+		You can have it cache **all** the arguments by value with `JSON.stringify`, if they are compatible:
+
+		```js
+		mem(function_, {cacheKey: JSON.stringify});
+		```
+
+		Or you can use a more full-featured serializer like [serialize-javascript](https://github.com/yahoo/serialize-javascript) to add support for `RegExp`, `Date` and so on.
+
+		```js
+		const serializeJavascript = require('serialize-javascript');
+
+		mem(function_, {cacheKey: serializeJavascript});
+		```
+
+		@default arguments_ => arguments_[0]
+		@example arguments_ => JSON.stringify(arguments_)
 		*/
 		readonly cacheKey?: (arguments: ArgumentsType) => CacheKeyType;
 
 		/**
-		Use a different cache storage. You could for example use a `WeakMap` instead or [`quick-lru`](https://github.com/sindresorhus/quick-lru) for a LRU cache.
+		Use a different cache storage. Must implement the following methods: `.has(key)`, `.get(key)`, `.set(key, value)`, `.delete(key)`, and optionally `.clear()`. You could for example use a `WeakMap` instead or [`quick-lru`](https://github.com/sindresorhus/quick-lru) for a LRU cache.
 
 		@default new Map()
+		@example new WeakMap()
 		*/
 		readonly cache?: CacheStorage<CacheKeyType, {data: ReturnType; maxAge: number}>;
 	}
--- a/index.d.ts
+++ b/index.d.ts
@@ -24,7 +24,7 @@ declare namespace mem {
 
 		You could for example change it to only cache on the first argument `x => JSON.stringify(x)`.
 		*/
-		readonly cacheKey?: (...arguments: ArgumentsType) => CacheKeyType;
+		readonly cacheKey?: (arguments: ArgumentsType) => CacheKeyType;
 
 		/**
 		Use a different cache storage. You could for example use a `WeakMap` instead or [`quick-lru`](https://github.com/sindresorhus/quick-lru) for a LRU cache.
@@ -32,13 +32,6 @@ declare namespace mem {
 		@default new Map()
 		*/
 		readonly cache?: CacheStorage<CacheKeyType, {data: ReturnType; maxAge: number}>;
-
-		/**
-		Cache rejected promises.
-
-		@default true
-		*/
-		readonly cachePromiseRejection?: boolean;
 	}
 }
 
--- a/readme.md
+++ b/readme.md
@@ -24,15 +24,19 @@ const memoized = mem(counter);
 memoized('foo');
 //=> 1
 
-// Cached as it's the same arguments
+// Cached as it's the same argument
 memoized('foo');
 //=> 1
 
-// Not cached anymore as the arguments changed
+// Not cached anymore as the argument changed
 memoized('bar');
 //=> 2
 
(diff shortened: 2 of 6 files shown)


readme.md
```diff
@@ -101,9 +105,20 @@ Milliseconds until the cache expires.
 
 Type: `Function`
 
-Determines the cache key for storing the result based on the function arguments. By default, if there's only one argument and it's a [primitive](https://developer.mozilla.org/en-US/docs/Glossary/Primitive), it's used directly as a key (if it's a `function`, its reference will be used as key), otherwise it's all the function arguments JSON stringified as an array.
+Determines the cache key for storing the result based on the function arguments. By default, only the first argument is used, as is, as the `key` of the `cache` `Map`.
+
+You change it to cache **all** the arguments by value with `JSON.stringify`, if they are compatible:
```
fregante:
```suggestion
You can have it cache **all** the arguments by value with `JSON.stringify`, if they are compatible:
```

readme.md
```diff
@@ -101,9 +105,20 @@ Milliseconds until the cache expires.
 
 Type: `Function`
 
-Determines the cache key for storing the result based on the function arguments. By default, if there's only one argument and it's a [primitive](https://developer.mozilla.org/en-US/docs/Glossary/Primitive), it's used directly as a key (if it's a `function`, its reference will be used as key), otherwise it's all the function arguments JSON stringified as an array.
+Determines the cache key for storing the result based on the function arguments. By default, only the first argument is used, as is, as the `key` of the `cache` `Map`.
```
fregante:
```suggestion
Determines the cache key for storing the result based on the function arguments. By default, **only the first argument is considered**.

A `cacheKey` function can return any type supported by `Map` (or whatever structure you use in the `cache` option).
```

readme.md
```diff
@@ -101,9 +105,20 @@ Milliseconds until the cache expires.
 
 Type: `Function`
```
fregante:
```suggestion
Type: `Function`
Default: `_arguments => _arguments[0]`
Example: `_arguments => JSON.stringify(_arguments)`
```

index.d.ts
```diff
@@ -24,21 +24,14 @@ declare namespace mem {
```
sindresorhus: You need to update the description here too. Just use everything used in the readme.

readme.md
```diff
@@ -100,10 +104,25 @@ Milliseconds until the cache expires.
...
+
+Determines the cache key for storing the result based on the function arguments. By default, **only the first argument is considered**.
+
+A `cacheKey` function can return any type supported by `Map` (or whatever structure you use in the `cache` option).
 
-Determines the cache key for storing the result based on the function arguments. By default, if there's only one argument and it's a [primitive](https://developer.mozilla.org/en-US/docs/Glossary/Primitive), it's used directly as a key (if it's a `function`, its reference will be used as key), otherwise it's all the function arguments JSON stringified as an array.
+You can have it cache **all** the arguments by value with `JSON.stringify`, if they are compatible:
+
+```js
+mem(function_, {cacheKey: JSON.stringify})
```
sindresorhus:
```suggestion
mem(function_, {cacheKey: JSON.stringify});
```

readme.md
```diff
@@ -100,10 +104,25 @@ Milliseconds until the cache expires.
...
+```
 
-You could for example change it to only cache on the first argument `x => JSON.stringify(x)`.
+Or you can use a more full-featured serializer like [serialize-javascript](https://github.com/yahoo/serialize-javascript) to add support for `RegExp`, `Date` and so on.
+
+```js
+const serializeJavascript = require('serialize-javascript');
+mem(function_, {cacheKey: serializeJavascript})
```
sindresorhus:
```suggestion

mem(function_, {cacheKey: serializeJavascript});
```

sindresorhus:
I agree, this makes a lot of sense. Thanks for getting it done.

readme.md
```diff
@@ -100,10 +104,25 @@ Milliseconds until the cache expires.
 ##### cacheKey
 
 Type: `Function`
+Default: `_arguments => _arguments[0]`
+Example: `_arguments => JSON.stringify(_arguments)`
+
```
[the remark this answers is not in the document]
fregante: I added that note in the main description in the readme + in the `cacheKey` description in JSDoc.

PR #47, opened 2019-11-15
Expose a `has` method to check the `cacheStore`

roblav96:
With the recent `v6.0.0` update, calling `clear` on a non-memoized function will throw.

## Problem
There's no way to check the `cacheStore` since it's not an exported variable.

## Solution
Expose a `has` method to check the `cacheStore` if function is memoized.

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,6 +43,10 @@ const mem = (fn, {
 
 module.exports = mem;
 
+module.exports.has = fn => {
+	return cacheStore.has(fn);
+}
+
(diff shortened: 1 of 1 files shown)


fregante:
Why would you call `clear` on a function that wasn't memoized? It's meant to be used exclusively on functions you passed through `mem(fn)`.

Especially if you're planning to clear the cache on a function, I would think that the code is pretty clear that said function was memoized. No?

PR #48, opened 2019-11-16
Support overloaded functions (TypeScript types)

### LINKED ISSUE #46: Automatically get types from memoized function
It looks like the types of `fn` are lost but something like this seems to keep them correctly, including any overloads:

```ts
const pMemoize: <T = VoidFunction>(fn: T) => T;
```

Same for `mem`. Solution:

```ts
declare module 'mem' {
	function mem<T = VoidFunction>(fn: T): T;
	export = mem;
}
```

fregante:
Fixes #46 

The problem was that the input function was being broken into `...Arguments` and `ReturnType` and then recomposed as `mem`’s return. This apparently picked up just one of the overloads and discarded the rest.

I tested this in Refined GitHub and it fixed my problems: https://github.com/sindresorhus/refined-github/pull/2545/files#diff-98a7886093e2e560968c34d6c4707b22L24-L29


fregante:
cc @BendingBender to possibly double-check

PR #51, opened 2020-01-02
Add DeepWeakMap tip to the readme

futpib:


readme.md
```diff
@@ -174,6 +174,31 @@ const memGot = mem(got, {cache});
 })();
 ```
 
+### WeakMap cache with multiple arguments
+
+If you want to memoize a function of multiple object arguments, you can use [deep-weak-map](https://github.com/futpib/deep-weak-map) as a replacement for the default cache.
```
sindresorhus:
```suggestion
If you want to memoize a function with multiple object arguments, you can use [deep-weak-map](https://github.com/futpib/deep-weak-map) as a replacement for the default cache.
```

sindresorhus: Can you include something about what problem your module solves? Like, how does it solve the problem of memoizing multiple object arguments? And how does it do it differently from, for example, `JSON.stringify()` or `serialize-javascript` which are mentioned elsewhere in the docs.

readme.md
```diff
@@ -174,6 +174,31 @@ const memGot = mem(got, {cache});
 })();
 ```
 
+### WeakMap cache with multiple arguments
```
fregante: I feel like this shouldn't be here. The `cacheKey` section already mentions:

> You can have it cache **all** the arguments by value with JSON.stringify, if they are compatible:

So you should extend that, which a minimal example like the 2 existing ones, clarifying _when_ you'd want to use this over a regular stringifier.

futpib: I tried to do this when addressing this comment https://github.com/sindresorhus/mem/pull/51#discussion_r362447271, is it better now?

fregante: No, it needs to be in the `cacheKey` section **and** the example needs to be trimmed to match the other two examples in that description. We don't need a full example with unrelated requires, it makes it harder to follow.

futpib: Are you sure about this? This use case covers both `cache` and `cacheKey` options.

fregante: Indeed. Then all examples should be moved outside `cacheKey`. They're all about different ways to use the cache, so they should be listed together and clearly.

I think the section should be called: **Using the right cache for your function** or something along those lines. In it, probably you can start with that's in `cacheKey` now and then continue with another minimal example using DeepWeakMap

fregante:
You're not following my reviews, so why ask for them?

fregante:
@sindresorhus: still waiting for https://github.com/sindresorhus/mem/pull/51#discussion_r369671208

Currently the usage of `cache`/`cacheKey` is split in two areas, but I think they should be merged into a new section called `Using the right cache for your function`

PR #55, opened 2020-04-06
Improve documentation about custom keys

fregante:
Replaces and closes #51 

Preview: https://github.com/sindresorhus/mem/blob/advanced-cache/readme.md#mem-


sindresorhus:
Very nice! This makes it totally clear 👌

PR #56, opened 2020-04-10
Automatically use WeakMap when possible

fregante:
`mem` doesn't **_need_** `Map`, it's just that `WeakMap` doesn't support most primitives. If it could, it should _only_ use `WeakMap`

Proposal: use `WeakMap`...

- if `cache` isn't provided by the user
- when the current `key` is supported by `WeakMap`

This is automatic and silent.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -193,9 +193,9 @@ Refer to the [caching strategies](#caching-strategy) section for more informatio
 ##### cache
 
 Type: `object`\
-Default: `new Map()`
+Default: `new Map()`, but it also intelligently uses `new WeakMap()` whenevever possible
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
It needs to be documented though: https://github.com/sindresorhus/mem#cache

readme.md
```diff
@@ -193,9 +193,9 @@ Refer to the [caching strategies](#caching-strategy) section for more informatio
 ##### cache
 
 Type: `object`\
-Default: `new Map()`
+Default: `new Map()`, but it also intelligently uses `new WeakMap()` whenevever possible
 
-Use a different cache storage. Must implement the following methods: `.has(key)`, `.get(key)`, `.set(key, value)`, `.delete(key)`, and optionally `.clear()`. You could for example use a `WeakMap` instead or [`quick-lru`](https://github.com/sindresorhus/quick-lru) for a LRU cache.
+Use a different cache storage. Must implement the following methods: `.has(key)`, `.get(key)`, `.set(key, value)`, `.delete(key)`, and optionally `.clear()`. You could for example use [`quick-lru`](https://github.com/sindresorhus/quick-lru) for a LRU cache.
```
fregante, line 198: 👌❓

PR #58, opened 2020-05-16
Revert "Automatically use WeakMap when possible"

fregante:
Great in theory, but the cache can't be `.clear()`ed anymore.

PR #59, opened 2020-05-16
Throw on clear(fn) if fn’s cache can't be cleared

fregante:
Breaking change, to be released after #58 (which is a patch for the current version)

What's the point of calling `clear(fn)` if it does nothing? This just causes silent errors.

Regarding throwing errors:

- I can **drop** the explicit `Can't clear a function that was not memoized!` error since now it throws anyway, OR
- I can **add** an `unclearable cache` error

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -141,12 +141,14 @@ Clear all cached data of a memoized function.
 @param fn - Memoized function.
 */
 mem.clear = (fn: (...arguments_: any[]) => any): void => {
-	if (!cacheStore.has(fn)) {
-		throw new Error('Can\'t clear a function that was not memoized!');
+	const cache = cacheStore.get(fn);
+	if (!cache) {
+		throw new TypeError('Can\'t clear a function that was not memoized!');
 	}
 
-	const cache = cacheStore.get(fn);
-	if (typeof cache.clear === 'function') {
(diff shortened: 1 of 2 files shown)


sindresorhus:
When would `cacheStore.get(fn).clear();` throw?

fregante:
It throws if `fn` is missing: `cacheStore.get(fn) is undefined`

It throws if `fn`’s Map doesn’t have clear: `cacheStore.get(fn).clear is not a function`

sindresorhus:
> I can add an unclearable cache error

👍

sindresorhus:
Is it really worth doing another breaking release just for this? I suggest releasing what's in master as a minor release and then merging this, but not release it until we need to do another release.

fregante:
You could also keep this PR open until there’s another reason to release a breaking version.

sindresorhus:
Yeah, we can do that.

opSSO:
Mem(rick.oem.mod)
Dmp(user.content.mod)
Limit(TEALS.remote.cat)
API(cockran.teals.op)

fregante:
> I can **add** an `unclearable cache` error

I pushed a change based on the latest version and added a test. This is now ready to be merged

Issue #60, opened 2020-08-20
Add function decorator


Richienb:
When function decorators become stable, we should use them to allow simplifying mem calls:

```js
const mem = require("mem")

@mem.Decorator()
function helloWorld() {
	//...
}
```

sindresorhus:
Would be nice. The decorator proposal has stalled for years though, so this might not happen anytime soon. I don't want to add it until there's support in at least the latest Node.js version.

fregante:
Why not? TypeScript supports them. Worst case scenario, you drop them or replace them in a future breaking release, but until then we can use them.

Currently using a simplified version on https://github.com/fregante/webext-options-sync-per-domain/blob/409d408150b8d7dfdd536204437e3b24f9128872/index.ts#L72

sindresorhus:
Last time I looked into decorators, TS implemented an obsolete version of the spec. That might have changed though. Anyway, I would merge a good PR for this, since there's interest.

PR #61, opened 2020-08-20
Add `.decorator()` method

Richienb:
Fixes: #60 

// @sindresorhus @fregante


sindresorhus:
It should be clearly documented that it's only for TypeScript consumers.

I'm also not sure about the "mixin" naming. "mixin" is a pretty overloaded term and it doesn't exactly fit here.

Richienb:
@sindresorhus > I'm also not sure about the "mixin" naming. 

See [how it's done in emittery](https://github.com/sindresorhus/emittery#emitterymixinemitterypropertyname-methodnames).

test.js
```diff
@@ -194,6 +210,6 @@ test('prototype support', t => {
 
 test('mem.clear() throws when called with a plain function', t => {
 	t.throws(() => {
-		mem.clear(() => {});
+		mem.clear(() => { });
```
sindresorhus, line 213: Can you disable Prettier for non-owned projects? I don't want to have to comment on this on every PR.

index.js
```diff
@@ -44,6 +44,16 @@ const mem = (fn, {
 
 module.exports = mem;
 
+module.exports.mixin = options => (target, propertyKey, descriptor) => {
+	if (typeof target !== 'function') {
+		throw new TypeError('`target` must be function');
```
sindresorhus, line 49:
```suggestion
		throw new TypeError('`target` must be a function');
```

test.js
Richienb: Prettier is already off. I don't get how vscode is still doing this.

index.d.ts
```diff
@@ -95,6 +95,37 @@ declare const mem: {
...
+	@example
+	```
+	const mem = require('mem');
+
+	class Example {
+		constructor() {
+			this.i = 0;
+		}
+
+		@mem.mixin();
```
sindresorhus, line 110: From looking at https://www.typescriptlang.org/docs/handbook/decorators.html, it seems decorators can be without the parens. Needs to be tested though.

sindresorhus:
> See how it's done in emittery.

It's called that there because the decorator actually "mixes in". That's not about the decorator itself.

> What should we call it instead?

Any suggestions?

test.js
sindresorhus: Try disabling all plugins and then enable them one-by-one until it happens again.

index.d.ts
Richienb: We want to use the return value of `mem.mixin()`, not the function itself.

Richienb:
@sindresorhus `.decorator()`?

test.js
Richienb: Found it in the built-in formatter options:

JavaScript › Format: Insert Space After Opening And Before Closing Nonempty Braces - Defines space handling after opening and before closing non-empty braces. Requires using TypeScript 2.3.0 or newer in the workspace.

index.d.ts
```diff
@@ -52,7 +52,7 @@ declare namespace mem {
 		@default new Map()
 		@example new WeakMap()
 		*/
-		readonly cache?: CacheStorage<CacheKeyType, {data: ReturnType; maxAge: number}>;
+		readonly cache?: CacheStorage<CacheKeyType, { data: ReturnType; maxAge: number }>;
```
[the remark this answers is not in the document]
fregante: This isn’t fixed. There are still unrelated whitespace changes

test.js
```diff
@@ -163,6 +163,22 @@ test('preserves the original function name', t => {
...
+		constructor() {
+			this.i = 0;
+		}
+
+		counter() {
+			return ++this.i;
+		}
+	}
+
+	mem.decorator()(TestClass.prototype.counter, 'counter', TestClass.prototype);
```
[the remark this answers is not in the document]
fregante: Whatever we support, we should write exact tests for it. The less real the tests, the less useful they are.

Since the implementations are different (https://github.com/babel/babel/issues/8864), if we support both we should also test both.

For the sake of simplicity (implementation, documentation, and testing), I'd suggest supporting TypeScript’s only.

index.d.ts
```diff
@@ -90,11 +90,42 @@ declare const mem: {
...
 		CacheKeyType,
 		FunctionToMemoize = (...arguments: ArgumentsType) => ReturnType
-	>(
+		>(
 		fn: FunctionToMemoize,
 		options?: mem.Options<ArgumentsType, CacheKeyType, ReturnType>
 	): FunctionToMemoize;
 
+	/**
+	Returns a decorator which memoizes the function provided.
```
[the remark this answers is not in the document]
sindresorhus: No

test.js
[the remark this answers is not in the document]
sindresorhus: I only want to support the TypeScript implementation.

sindresorhus:
@Richienb Bump

Richienb:
@sindresorhus Since this feature is currently only supported by Typescript, it's best if https://github.com/sindresorhus/mem/issues/62 is merged so I can implement proper tests.

fregante:
Can you rebase this PR onto `master`? You can move the decorator test to its own tests/decorator.ts file

Richienb:
@fregante Done!

fregante:
You’ll have to rebase the branch and make the tests pass. Currently I’m seeing changes that are already on master. This PR should only be of 2/3 files

fregante:
Proper TypeScript tests were merged so you can just append them to `test.ts` and make sure that the build passes.

index.js
```diff
@@ -44,6 +44,16 @@ const mem = (fn, {
 
 module.exports = mem;
 
+module.exports.mixin = options => (target, propertyKey, descriptor) => {
+	if (typeof target !== 'function') {
+		throw new TypeError('`target` must be a function');
+	}
+
+	descriptor.value = mem(target[propertyKey], options);
```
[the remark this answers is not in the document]
fregante: Please add a test for this, for example:

```ts
class Foo {
  i = 0;
  @mem.decorator();
  count() {
    i++;
  }
}

const a = new Foo();
t.is(a.i, 0);
a.count();
t.is(a.i, 1);
a.count();
t.is(a.i, 1); // Memoized 👌


const b = new Foo();
t.is(b.i, 0);
a.count();
t.is(b.i, 1); // count() was memoized across instances 👌
```

fregante:
Ping if you have time for this

Richienb:
Where did the build statuses go?

fregante:
You’ll have to merge master into this because Travis is gone

fregante:
Sorry I thought GHA had already been enabled here

sindresorhus:
https://github.com/sindresorhus/mem/commit/b1c65cda18c06271b0e70db00c804f38b6a326ac

Richienb:
@sindresorhus Are you at the point of switching to Github Actions for all of your projects?

sindresorhus:
@Richienb Still interested in finishing this? I noticed it's still marked as a draft PR and tests are failing.

Richienb:
Yes - most the errors look to be TypeScript compilation errors so I'll get to fixing them

index.ts
```diff
@@ -140,6 +140,53 @@ const mem = <
...
+}
+```
+*/
+mem.decorator = <
+	FunctionToMemoize extends AnyFunction,
+	CacheKeyType
+>(
+	options: Options<FunctionToMemoize, CacheKeyType> = {}
+) => (
+	target: any,
```
Richienb, line 179: I don't think there's a way to avoid this being `any`.

index.ts
```diff
@@ -140,6 +140,53 @@ const mem = <
...
+	propertyKey: string,
+	descriptor: PropertyDescriptor
+): void => {
+	if (typeof target[propertyKey] !== 'function') {
```
Richienb, line 183: Would it make sense to assign `target[propertyKey]` to a seperate variable to prevent double lookup or is that just a pointless microoptimisation?

index.ts
```diff
@@ -140,6 +140,53 @@ const mem = <
...
+		throw new TypeError('Decorated value is not a function');
```
sindresorhus, line 184:
```suggestion
		throw new TypeError('The decorated value must be a function');
```

sindresorhus (changes requested):

index.ts
Richienb: `target` refers to the object that the function lives in. `const input: FunctionToMemoize = target[propertyKey]`.

sindresorhus: Shouldn't it at least be `object` then to ensure `undefined` and `null` are not the target? Also, I generally prefer `unknown` over `any` whenever possible.

sindresorhus:
https://github.com/sindresorhus/mem/pull/61#discussion_r596805463

index.ts
Richienb: `object` isn't allowed by `xo` which wants it to be `Record<string, unknown>` which doesn't work either because a class is not an plain object.

sindresorhus: Not using `object` is just a general rule. You shouldn't use `any` either, ideally. But sometimes we need exceptions. Just use an `eslint-ignore` comment.

Richienb: As soon as I change `any` to `object`, the compiler throws because [`{}` can't be indexed by a `string`](https://stackoverflow.com/questions/48382306/typescript-how-skip-define-type-on-multiple-generics-class).

If I try to use generics, the compiler is instead unable to resolve a target properly so it always falls back to `unknown`.

Issue #50, opened 2019-12-04
[TypeScript] cacheKey arguments array isn't automatically typed


dylang:
In v5, the cacheKey arguments were automatically typed based on the method signature. In v6, the cacheKey arguments extend `unknown` instead of getting them from the original function.

Example error:

```ts
const example = mem(
  (str: string, strArray: string[]) => { /* do stuff */  },
  {
    // Typescript error on strArray: 
    // TS2461: Type 'unknown' is not an array type.
    cacheKey: ([str, strArray]) => [str, ...strArray].join()
    //                                   ^^^^^^^^^^^ <- Example error
  }
);
```

Workaround:

```ts
const example = mem(
  (str: string, strArray: string[]) => { /* do stuff */  },
  {
    // Repeat the types of the arguments inside of [ ].
    cacheKey: ([str, strArray]: [string, string[]]) => [str, ...strArray].join()
    //                        ^^^^^^^^^^^^^^^^^^^^ <- Example workaround
  }
);
```

I'm not sure if this can be fixed using generics, so it might just be some documentation for those confused when upgrading.

PR #63, opened 2020-09-29
Infer `cacheKey` arguments type

wzhkobe2408:
issue link: https://github.com/sindresorhus/mem/issues/50

solution: use`Parameters<typeof fn>` to infer type of cacheKey arguments

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,7 +8,7 @@ declare namespace mem {
 	}
 
 	interface Options<
-		ArgumentsType extends any[],
+		ArgumentsType extends unknown[],
 		CacheKeyType,
 		ReturnType
 	> {
@@ -86,13 +86,13 @@ declare const mem: {
 	```
 	*/
 	<
-		ArgumentsType extends any[],
+		ArgumentsType extends unknown[],
 		ReturnType,
 		CacheKeyType,
-		FunctionToMemoize extends (...arguments: ArgumentsType) => ReturnType
+		FunctionToMemoize = (...arguments: ArgumentsType) => ReturnType
 	>(
 		fn: FunctionToMemoize,
-		options?: mem.Options<Parameters<typeof fn>, CacheKeyType, ReturnType>
+		options?: mem.Options<FunctionToMemoize extends (...arguments: infer P) => ReturnType ? P : never, CacheKeyType, ReturnType>
 	): FunctionToMemoize;
 
 	/**
@@ -100,7 +100,7 @@ declare const mem: {
 
 	@param fn - Memoized function.
 	*/
-	clear<ArgumentsType extends any[], ReturnType>(
+	clear<ArgumentsType extends unknown[], ReturnType>(
 		fn: (...arguments: ArgumentsType) => ReturnType
 	): void;
 };
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,7 +8,7 @@ declare namespace mem {
 	}
 
 	interface Options<
-		ArgumentsType extends unknown[],
+		ArgumentsType extends any[],
 		CacheKeyType,
 		ReturnType
 	> {
@@ -86,13 +86,13 @@ declare const mem: {
 	```
 	*/
 	<
-		ArgumentsType extends unknown[],
+		ArgumentsType extends any[],
 		ReturnType,
 		CacheKeyType,
-		FunctionToMemoize = (...arguments: ArgumentsType) => ReturnType
+		FunctionToMemoize extends (...arguments: ArgumentsType) => ReturnType
 	>(
 		fn: FunctionToMemoize,
-		options?: mem.Options<FunctionToMemoize extends (...arguments: infer P) => ReturnType ? P : never, CacheKeyType, ReturnType>
+		options?: mem.Options<Parameters<FunctionToMemoize>, CacheKeyType, ReturnType>
 	): FunctionToMemoize;
 
 	/**
@@ -100,7 +100,7 @@ declare const mem: {
 
 	@param fn - Memoized function.
 	*/
-	clear<ArgumentsType extends unknown[], ReturnType>(
+	clear<ArgumentsType extends any[], ReturnType>(
 		fn: (...arguments: ArgumentsType) => ReturnType
 	): void;
 };
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,7 +8,7 @@ declare namespace mem {
 	}
 
 	interface Options<
-		ArgumentsType extends unknown[],
+		ArgumentsType extends any[],
 		CacheKeyType,
 		ReturnType
 	> {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add a type test to index.test-d.ts?

fregante:
@sindresorhus probably a breaking change to be released with #59

wzhkobe2408:
@sindresorhus  It seems like when `ArgumentsType extends any[]`, type test will pass, but with `ArgumentsType extends unknow[]`, type test will fail

wzhkobe2408:
@sindresorhus  There are some problem, i will try to figure it out, if i can, i will add a type test

sindresorhus:
@wzhkobe2408 Yes, it needs to be `any` for variables to be able to spread and not just inline literals (I didn't know this when writing the first definitions here).

wzhkobe2408:
@sindresorhus  I just found out that
```
	<
		ArgumentsType extends unknown[],
		ReturnType,
		CacheKeyType,
		FunctionToMemoize = (...arguments: ArgumentsType) => ReturnType
	>(
		fn: FunctionToMemoize,
		options?: mem.Options<FunctionToMemoize extends (...arguments: infer P) => ReturnType ? P : never, CacheKeyType, ReturnType>
	): FunctionToMemoize;
```

with `FunctionToMemoize extends (...arguments: infer P) => ReturnType ? P : never`, you don't need to change unknow to any

wzhkobe2408:
@sindresorhus  But is strange when i use typescript inner helper Parameters type, i will cause error

wzhkobe2408:
@sindresorhus   I think it fix the issue 😝

sindresorhus:
I would prefer the `any` approach. The types are simpler that way. And more consistent with all my other projects. I don't like using `any` either, but it's the blessed way to do argument spreading.

wzhkobe2408:
Use `any` or u`nknow` doesn't effect the way to infer cacheKey arguments type, eventually programmer will get correct type, so they know what to do with cacheKey arguments. You can continue to commit

sindresorhus:
> Use any or unknow doesn't effect the way to infer cacheKey arguments type

I explained what `any` does differently above. I'm not sure what you're arguing.

wzhkobe2408:
> > Use any or unknow doesn't effect the way to infer cacheKey arguments type
> 
> I explained what `any` does differently above. I'm not sure what you're arguing.

ok, I understand now

sindresorhus:
> Can you add a type test to index.test-d.ts?

owenallenaz:
I'm facing a similar issue with the cacheKey being out of sync, but I think in general the tyepscript signatures could be made much simpler... I think. I'm currently hacking around trying to see if I can figure out how to do it.

owenallenaz:
I've looked at this for awhile now, and I'm utterly lost. Either I'm missing some knowledge about how TS generics work, or some of this just doesn't make sense to me. At https://github.com/sindresorhus/mem/blob/master/index.d.ts#L88 there is a line

```
<
	ArgumentsType extends unknown[],
	ReturnType,
	CacheKeyType,
	FunctionToMemoize = (...arguments: ArgumentsType) => ReturnType
>(
	fn: FunctionToMemoize,
	options?: mem.Options<ArgumentsType, CacheKeyType, ReturnType>
): FunctionToMemoize;
```

There are a few elements of that signature that don't make sense to me. So either I'm off on my understanding or those signatures don't make sense. The generic definition has definitions for 4 params, but `mem` ultimately only takes 2 parameters. It would seem that `CacheKeyType` and `FunctionToMemoize` will always be unusable and effectively undefined. Is there a simplified example you can point to which illustrates how that concept works?  Another portion of that invocation is the pattern of `FunctionToMemoize = ...` in the generic definition. I'm lost on that too. In example if you look at a simplified example in the TS playground https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABFApgZygHgIIBpEBCiAvImCALYBGKATgHwAUAhgFyLYCU7RA3gFCIhiWiighaSAIwAGANz8AvkA, you can see that the assignment of a generic param to a simple type doesn't actually do anything.

The second issue is that at it's core the `options` object has an eccentricity where it appears we're attempting to have sibling keys infer based on each other, which feels impossible. In example, the cache.get(key) function depends on the ReturnType of cacheKey, but cacheKey and cache are siblings.

Am I missing some functionality of TS here? If not, then it seems like these signatures could be greatly simplified, if I am mistaken then a simple example should set me straight and I should be able to fix the outlined above issue that even led me here. Thanks.

fregante:
The types are complex because:

- `mem(fn)` should return the same type as `fn`
- the Map defined in `cache` should store the same **return value** of `fn`
- the `cacheKey` function should receive the same **arguments** of `fn`

I think that if someone converts `mem` to TypeScript, the types might be easier to figure out and maintain (because manual `.d.ts` files are not automatically checked against the original `.js` file)

This would be the right starting point for the configuration of a 100% TypeScript project: https://github.com/sindresorhus/got/blob/c31366b8c279c3cdab31ac5ae91d72ce68280980/package.json

fregante:
> because manual `.d.ts` files are not automatically checked against the original `.js` file

In fact they're wrong in multiple places. WIP

owenallenaz:
I don't see any of this as a problem that requires converting mem to TypeScript. In fact one of the essential goals and good things about TS is that it plays perfectly with JS/TS code intermixed. `mem` is perfectly valid for what it is, there's just some fixable bugs in the TS definitions, easily solved. The issue I have, as outlined in my previous statement is that the typescript definitions don't appear to make sense (again, unless I'm missing something). You'd have the exact same issues if it was written in typescript directly. In example, you can reproduce all of the problems I outline in the typescript playground, which is where I've been working.

If you happen to know how to have `cache` and `cacheKey` be siblings and yet infer type information from one to the other, please show me.

fregante:
> fixable bugs in the TS definitions, easily solved

Right, but the definition is wrong, and the fact that it's detached simply makes it less-easily maintainable because we can introduce bugs without even realizing it.

One example is that `mem.clear`’s types are wrong and there's nothing that verifies them:
https://github.com/sindresorhus/mem/blob/807d7c0ef27c9d76fe682103ea39e6763abcf051/index.d.ts#L103-L105

`ArgumentsType` has no meaning here because it's only defined when calling `mem()`, but `.clear` is not part of that.

In short: it's easier to mess up types when you don't verify them internally by writing pure TS and we totally do, over and over, mess them up. Technically you could place `export = (a: any) => any` in `index.d.ts` and the package publishes just fine.

fregante:
I tried the types from this PR into #64, but they didn't affect the output.

@wzhkobe2408 I also tried using your code from https://github.com/sindresorhus/mem/pull/63#issuecomment-700605617 but wasn't able to make it work. You can open a PR _onto my PR_ if you'd like

fregante:
@wzhkobe2408 I spoke too soon, this _does_ appear to fix this issue.

I'm not sure why I wasn't able to port it over to #64

owenallenaz:
@fregante I took what you started with and integrated it with my simplifications and it appears to be working. I have submitted a separate PR with my variant that simplifies and fixes the few bugs that were outstanding in the TS signatures, I think.

https://github.com/sindresorhus/mem/pull/65

fregante:
Closing this in favor of #64 and #66 

Thanks @wzhkobe2408 for sparking this 💥 ^_^

PR #64, opened 2020-10-01
Convert to TypeScript

### LINKED ISSUE #62: Convert tests to TypeScript
To facilitate #60/#61, the tests be converted to TypeScript.

Refer to: https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md

fregante:
Related: #63 
Fixes #62 

Background about the types: https://github.com/sindresorhus/mem/pull/63#issuecomment-701735275

<details>
<summary>Outdated details</summary>

I tested this on Refined GitHub. TypeScript worked fine, but XO started complaining about:

```
  source/features/view-markdown-source.tsx:83:2
  ✖  83:2   This assertion is unnecessary since it does not change the type of the expression.  @typescript-eslint/no-unnecessary-type-assertion
  ✖  84:9   This assertion is unnecessary since it does not change the type of the expression.  @typescript-eslint/no-unnecessary-type-assertion

  source/features/restore-file.tsx:53:17
  ✖  53:17  This assertion is unnecessary since it does not change the type of the expression.  @typescript-eslint/no-unnecessary-type-assertion
```

On code like [`dom! = memoizedFunction()`](source/features/view-markdown-source.tsx), but dropping the `!` causes TypeScript to fail.

It's probably breaking,  let's merge #59 first before publishing a new version

</details>


index.ts
```diff
@@ -0,0 +1,153 @@
...
+	maxAge
+}: Options<ArgumentsType, CacheKeyType, ReturnType> = {}): FunctionToMemoize => {
+	if (typeof maxAge === 'number') {
+		// @ts-expect-error
+		// TODO: remove after https://github.com/SamVerschueren/map-age-cleaner/issues/5
+		mapAgeCleaner(cache);
+	}
+
+	// @ts-expect-error
+	const memoized: FunctionToMemoize = function (this: any, ...arguments_: ArgumentsType): ReturnType {
```
fregante, line 108: Error:
```
Type '(this: any, ...arguments_: ArgumentsType) => ReturnType' is not assignable to type 'FunctionToMemoize'.
  '(this: any, ...arguments_: ArgumentsType) => ReturnType' is assignable to the constraint of type 'FunctionToMemoize', but 'FunctionToMemoize' could be instantiated with a different subtype of constraint '(...arguments_: ArgumentsType) => ReturnType'.ts(2322)
```

fregante:
@BendingBender @nickytonline could you take a look at the types if you're familiar with `mem`?

owenallenaz:
Lol... I was just working on merging my stuff in and was restoring the tsd file.

owenallenaz:
Alright PR created in #66 that includes the newer updates placed onto the TS branch. Ran into a bunch of awkward issues that took way longer than they should to correct. Mainly I didn't know you can prefix a var with `_` to cause it to not trigger unused var errors (which were everywhere in the tsd stuff). I also moved the tsd tests into a type_tests folder, this way we can get it out of dist and unwind some of the awkwardness you had to do to add it back in.

One thing that still seems off is that the dist folder contains the compiled test files. According to the ava docs (https://github.com/avajs/ava/blob/HEAD/docs/recipes/typescript.md) they mention using ts-node that way you don't have to precompile the test files and therefore they don't end up in dist.

fregante:
Precompiling tests seems to be the preferred way: https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md

In fact I copied the config from one of AVA's modules: https://github.com/avajs/cooperate/blob/1d82c6cb09dfdc426ce75c6d634a3c863b897af4/package.json#L16-L18

I also wish AVA could just read .ts files with ts-node, but I found that it introduces more problems.

nickytonline:
I’ll take a peek at this this week.

fregante:
- No tests had to be changed
- AVA tests pass
- Type tests pass
- Refined GitHub compiles without issue

For all intents and purposes, this PR doesn't bring any breaking changes and is a good starting point for further PRs.

For this reason, I'll merge it and types improvements can still follow before the next release, perhaps in #66

PR #65, opened 2020-10-01
Fixes a bug in cacheKey, simplifies, adds more robust testing to TS signatures

owenallenaz:
This corrects and adds more robust testing for a variety of flaws in the TS signatures for mem.

* The `cacheKey` function was receiving the improper signature. As the code is written the argument in the first position will be the totality of the arguments array so `cacheKey : function(args) {}` where `args = [arg0, arg1, arg2]`.  The previous code was using the magic `arguments` which resulted in `cacheKey : function(arg0, arg1) {}` which was incorrect.
* Using `Parameters` and `ReturnType` we can make use of some of the builtins that TS provides to simplify. 
* There wasn't a robust testing for the interaction between the `cacheKey` and `cache` in the way that the return type of `cacheKey` affects the incoming parameters received by `cache`. Added a complex test for that example.
* The `clear` method had a much more complicated than necessary signature. Effectively all that matter is that it takes a function, right?
* Pre-existing tests were erroneously declaring the Map as `Map<[string], ...>` that would indicate that the key in the Map is an array of strings, but that isn't the case for the default state of `mem`. If no cacheKey is specified the default is the first argument, and in that test the first argument is a string, so the default map state should be string.

Hit me up for any feedback or issues. Thanks.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,16 +1,22 @@
 declare namespace mem {
+	type AnyFunc = (...args: any) => any;
+	
+	interface CacheStorageContent<ValueType> {
+		data: ValueType
+		maxAge: number
+	}
+
 	interface CacheStorage<KeyType, ValueType> {
 		has(key: KeyType): boolean;
-		get(key: KeyType): ValueType | undefined;
-		set(key: KeyType, value: ValueType): void;
+		get(key: KeyType): CacheStorageContent<ValueType> | undefined;
+		set(key: KeyType, value: CacheStorageContent<ValueType>): void;
 		delete(key: KeyType): void;
 		clear?: () => void;
 	}
 
 	interface Options<
-		ArgumentsType extends unknown[],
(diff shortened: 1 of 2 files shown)


fregante:
Can you apply these changes onto my branch? I already converted everything to TypeScript, there’s no reason to go back

owenallenaz:
If @sindresorhus indicates that he plans on moving to a full TS variant than I'll happily integrate it in. I've looked at the other PRs and issues and I'm not certain he's signaled one way or the other if he's looking to move the whole project to native TS.

fregante:
@owenallenaz https://github.com/sindresorhus/mem/pull/64#event-3828819914

sindresorhus:
@owenallenaz Yes, @fregante and I have agreed to convert the whole package to TS. Would be great if you could target his branch.

fregante:
Duplicate of #66

PR #66, opened 2020-10-02
Fixes a bug in cacheKey, simplifies, adds more robust testing to TS signatures

owenallenaz:
Fixes https://github.com/sindresorhus/mem/issues/50
Merges into https://github.com/sindresorhus/mem/pull/64


package.json
```diff
@@ -69,5 +69,8 @@
 			"@typescript-eslint/no-var-requires": "off",
 			"@typescript-eslint/no-empty-function": "off"
 		}
+	},
+	"tsd": {
+		"directory": "type_tests"
```
sindresorhus, line 74: The convention is `test-d`. Then we don't have to specify it here. https://github.com/SamVerschueren/tsd/#test-directory

index.ts
```diff
@@ -1,20 +1,27 @@
 'use strict';
 import mimicFn = require('mimic-fn');
-import mapAgeCleaner = require('map-age-cleaner');
+import mapAgeCleaner from 'map-age-cleaner';
+
+type AnyFunc = (...args: any) => any;
```
sindresorhus, line 5:
```suggestion
type AnyFunction = (...args: any) => any;
```

sindresorhus (changes requested):

fregante:
There's still an error in `test.ts` that is supposedly due to #50: https://github.com/sindresorhus/mem/pull/64/files#diff-9d878e6c8076c7cdd71130118d1fc802R38

package.json
owenallenaz: Done.

owenallenaz:
Build fully passes per my last commit. I'm not seeing any errors in that test when I run my branch locally nor in Travis.

fregante:
> Build fully passes per my last commit
They build passes because the errors are ignored. Open the 2 links I posted

fregante:
Side note, type tests added by this PR are welcome, but are unrelated to #50 as well:

These tests also pass on `master` without change: https://github.com/sindresorhus/mem/commit/719fd10d72bbdfd03b01d0d6d8e0aa095b979952

owenallenaz:
So for those 2 tests, the fn being tested doesn't actually declare any args in it's definition, so in that case cacheKey will not know them, because they aren't declared. So using `any` in that situation is perfectly valid. Those tests aren't for testing the type checking either, so I see no harm in using `any` in that context.

So the second issue is that on `master` you are relying on `unknown[]` which will result in type tests always passing because anything can be `unknown`. An easy way to see this is in your example test https://github.com/sindresorhus/mem/commit/719fd10d72bbdfd03b01d0d6d8e0aa095b979952#diff-3ff567c951fb027d77a850244e0e6303R31 on line 31. If you change that line from 

```
cacheKey: (_args: [{key: string}]) => {
```

to

```
cacheKey: (_args: [{key: number}]) => {
```

You'll see, on master, the tests still pass. On my branch, they do not, because I am not relying on `unknown[]`. The reason for the more verbose test on cacheKey and cache is that, prior to that, there was no robust checking any the cascade from cacheKey's returnType was affecting the cache entity. It was functional, but not proven.

test.ts
```diff
@@ -142,8 +141,7 @@ test('cache option', t => {
 	const fixture = () => i++;
 	const memoized = mem(fixture, {
 		cache: new WeakMap(),
-		// @ts-expect-error Due to https://github.com/sindresorhus/mem/issues/50
-		cacheKey: ([firstArgument]) => firstArgument
+		cacheKey: ([firstArgument]: [any]) => firstArgument
```
fregante, line 144: I don't think this is the right type and this test was essentially wrong. There was never a `firstArgument` so this shouldn't pass. The type of cacheKey should be

```ts
(...arguments: Parameters<fixture>) = CacheKeyType
```

`Parameters<fixture>` is strictly `[]`, not `any[]`.

fregante: Live demo: https://www.typescriptlang.org/play?#code/DYUwLgBAlhC8EAYDcAoAxgewHYGdIDMoAPMAVwCcQ4IAKASjgD5oBqF1MATwAcqBhAIZoAFiADSITtRoA6OQPIBzUgFsQWMDgD6ALggAFBQLVgQ5HAB4uvDPgiESFEIwaxmArJ1SZckNENEJTj1BEXFJaVdmAGUwcigsRRoARjokCAB6DIgAeTEAQnRsPAh-MKCAJhCA8Kl4GgBtQnMwAEElVXUwAF0oiFj4xJpmvHblNQ00zOyAORyAFVyCoA

test-d/index.test-d.ts
```diff
@@ -0,0 +1,63 @@
...
+				set: (_key: Date, _value: { data: number; maxAge: number }) => {},
+				has: (_key: Date) => {
+					return true;
+				},
+				delete: (_key: Date) => {},
+				clear: () => {}
+			}
+		}
+	)
+);
```
fregante, line 50: Since so many things were changed at once in #64 and #66 it's best we split these PRs. Moving files around especially makes it hard to see what changed exactly.

Do you mind sending this `cacheKey` test separately? I'll let you know when you can rebase this PR

fregante:
Sorry for the mess of commits, I tried to resolve the conflicts and rebasing it to `master`

Here's the gist:

- changes to config files were dropped since I sorted it out in #64 
- changes to test.ts were dropped because the types in tests were actually meaningless: https://github.com/sindresorhus/mem/pull/68#issuecomment-704044169
- so that leaves index.ts and index.test-d.ts. The latter can also be dropped once you follow https://github.com/sindresorhus/mem/pull/66#discussion_r500014172 since the other changes were incorrect (you can see the comments I added inline)

fregante:
I finally figured out how to write type tests for #50: https://github.com/sindresorhus/mem/commit/837e393531fa26002f860ff038901427597009fd

You can use `expectType` anywhere, not just at the top level.

That test is now on `master`

test.ts
owenallenaz: I believe there are 2 flaws in what you're doing. First, in your TS playground example you have cacheKey receiving `(...arguments_: Parameters<typeof fixture>) => any;`. That isn't correct. cacheKey receives `(arguments_: Parameters<typeof fixture>) => any`. The first argument IS the array. The additional `...` causes it to flatten out. That was the flaw that was in `master` branch.

The other issue is that according to `mem` code we **always** have a firstArgument, because it's the array.

```
const memoized = function (this: any, ...arguments_) {
    const key = cacheKey ? cacheKey(arguments_) : arguments_[0];
```

In example in your js console if you run

```
function foo(...arguments_) { console.log(arguments_); }
console.log(foo())
// []
console.log(foo("bar"))
// ["bar"]
console.log(foo("bar", "baz"))
// ["bar", "baz"]
```

So then cacheKey will receive `[]`. In this circumstance we're effectively calling a function which we gave TS no type hinting. In the code of the test we're returning `String(firstArgument)` and that's totally valid because we're sometimes passing it an int, and sometimes a string, so we have to instruct TS that we know what we're doing and `any` it or `string | number`. The intent of this test isn't really to verify the TS integrity, it's to verify the caching, so I believe using `any` is totally valid.

owenallenaz:
I've followed up in a few of the above comments. But I'm not exactly sure what you're looking for at this point? Do I need to do anything further?

fregante:
Do you still want to fix #50? That's still outstanding.

Also this: https://github.com/sindresorhus/mem/pull/66#discussion_r500014172

test.ts
fregante: > The other issue is that according to `mem` code we **always** have a firstArgument, because it's the array.

That's slightly incorrect. It does have `arguments_`, not `[firstArgument]`. Notice that the latter is being destructured. `arguments_` will be the tuple `[]`, so there can't be a `[firstArgument]` in there.



> so I believe using `any` is totally valid.

For the reason I just explained, it's not. You can see this in the latest type tests I added, which are being ignored: https://github.com/sindresorhus/mem/commit/837e393531fa26002f860ff038901427597009fd

`any[]` and `[any]` can't be assigned to `[]`.

owenallenaz:
#50 is fixed in my branch. 

Regarding the complicated PRs, if you're moving to TS, I think the easiest thing would be to merge your TS variant which changes nothing (pure code migration, no functional changes). Then I can PR my fixes to master so then rather than it being one giant overhaul, it's now two distinct efforts. Your conversion + my fixes. You don't need to publish to NPM in between, but it gets the code on master and easier to review my smaller incremental changes?

fregante:
> to merge your TS variant

That's what I did already.

> I can PR my fixes to master

I rebased this PR already. You can open a new one or force push here. As you see fit.

> You don't need to publish to NPM in between

Yeah we're still waiting to figure this out.

> #50 is fixed in my branch.

It does appear to be 🎉 

```
❯ npx tsd

  test-d/index.test-d.ts:69:2
  ✖  69:2  Unused @ts-expect-error directive.  
  ✖  76:2  Unused @ts-expect-error directive.  

  2 errors
```

owenallenaz:
Starting on this guy again, will complete this morning.

sindresorhus:
https://github.com/sindresorhus/mem/pull/68 was merged now. Would be great if you could rebase from master.

owenallenaz:
Simplified clean PR at #69 . Closing this PR.

PR #68, opened 2020-10-06
Convert tests to TypeScript

fregante:
From #64, follows #67


fregante:
Heh, so I think that tests written in TypeScript are a bad idea.

Tests for JavaScript code try to throw all kinds of dirt at the module to see if it still works. TypeScript avoids this at runtime. You can't test _"let's see what happens with 2 arguments"_ and expect TypeScript to just let you do that.

In #64 and #66 I was mistakenly using `const x = require` instead of `import x = require`, which means **`mem` was always of type `any`.**

So here's what happens when I use the latter ⬇️

fregante:
You can see what the problem is:

---

The reason why I was converting the tests to TypeScript was https://github.com/sindresorhus/mem/pull/61#issuecomment-701070585

#61 will need its own TypeScript tests file, but for the main tests we have 2 options:

- leave them in `test.js` so that nobody expects them to be typed
- use the first commit here, where the test is "in TypeScript" but `mem` is `any`. This only makes sense if the first option is a mess to configured.

@sindresorhus @owenallenaz thoughts?

owenallenaz:
I do believe it's valuable to write tests in TS, it's a form of dog-fooding your own code. Half the exercise of writing tests is using your own stuff and finding the ways it's busted, possibly including the TS signatures, which may not uncover when writing the tests in JS (well, it might if you're using vscode or something with integrated TS mechanics). In one of your comments you mentioned how you can't test sending two args to a function that accepts one arg, but you absolutely can.

```
function oneArg(foo: string) {}

// valid, function will receive number 10
oneArg(10 as any)
//@ts-ignore
oneArg(10, "another");
```

In testing, we often want to use `any` and escaping out with `@ts-ignore` because we are explicitly testing bad code. Some people will say @ts-ignore and `any` are a code-smells, but I think that's mistaken for test files and the circumstances you outline. It's better to have ts working for 90% of a test file, and @ts-ignore/any your way around the other 10% then to basically turn TS off and move back to JS, since now we've effectively @ts-ignored an entire file, so if it's a smell to use it a little, it should be smellier to use it for the whole suite, no?

Generally the way I have approached my ts projects is I do the following:

* All code in TS
* Most tests in TS.
* TS tests are compiled hot by the `ts-jest` so that I don't have to build first to run the primary unit tests.
* Another, much, much smaller test suite that tests the main /dist/ or /build/ which verifies that it's exporting correctly. In example, in my main TS project, we have a very complex rollup system. In the js tests, I require the package in /dist/ and verify it rolled up correctly by just smoke testing a few mechanics. In example, this would catch the issue you outlined with the bundle being broken.
* So the "release" process is `npm test`, `npm build`, `npm test`, `npm publish`. The awkwardness is needing to run build between the two tests, but usually with the frequency that I'm hammering `npm test` I don't want every test run to be weighed down by a new build, especially if I'm doing an `.only()` and working within one test file, which ensures that `ts-jest` doesn't even hot compile the whole package.

sindresorhus:
I also think it's valuable to have the tests in TS.

sindresorhus:
> TypeScript avoids this at runtime.

I think you meant *compile-time*.

> You can't test "let's see what happens with 2 arguments" and expect TypeScript to just let you do that.

This is when it's ok to use a lot of `@ts-expect-error`.

fregante:
> you can't test sending two args to a function that accepts one arg, but you absolutely can
> ```ts
> //@ts-ignore
> ```

Did you open the Travis link? There are 54 type error which likely need to be ignored by TypeScript. I don't see much use in having TypeScript test if you're going to ignore 30% of a file. 

> it should be smellier to use it for the whole suite, no?

Absolutely not. If you're going to ignore 30% of your lines you're just giving anyone the license to ignore any line in a test file. This makes reviewing difficult:

_Is this line ignored because we're testing JavaScript use or because it's not valid in the first place?_



> I think you meant _compile-time_.

Yes.

fregante:
Feel free to merge the above. They are TypeScript tests, following your suggestions, but who knows if they're valid?

owenallenaz:
Another technique to avoid tests in ts is if it's the "unused variable" warning, simple prefix it with _ will cause ts to auto-ignore unused vars.

test.ts
```diff
@@ -126,30 +170,37 @@ test('maxAge items are deleted even if function throws', async t => {
...
 	t.is(cache.size, 1);
 	await delay(50);
+	// @ts-expect-error
 	t.is(memoized(1), 0);
 	await delay(200);
-	t.throws(() => memoized(1), {
-		message: 'failure'
-	});
+	// @ts-expect-error
+	t.throws(() => memoized(1), {message: 'failure'});
```
sindresorhus, line 183:
```suggestion
	t.throws(() => {
		memoized(1);
	}, {message: 'failure'});
```

test.ts
```diff
@@ -126,30 +170,37 @@ test('maxAge items are deleted even if function throws', async t => {
...
 	t.is(cache.size, 0);
 });
 
 test('cache option', t => {
 	let i = 0;
 	const fixture = () => i++;
 	const memoized = mem(fixture, {
+		// @ts-expect-error
```
sindresorhus, line 191: This should not be ignored though. The `cache` option should support WeakMap.

fregante: That's right. However #66 is already touching those exact types so perhaps it's best to merge this and have that PR fix any remaining issues.

If you prefer, I can drop this comment and have it fail for the PR to pick up.

test.ts
```diff
@@ -206,6 +259,7 @@ test('mem.clear() throws when called with a plain function', t => {
 test('mem.clear() throws when called on an unclearable cache', t => {
 	const fixture = () => 1;
 	const memoized = mem(fixture, {
+		// @ts-expect-error
```
fregante, line 262:
```suggestion
```

test.ts
fregante:
```suggestion
```

fregante (commented):
If you want to merge the PR while failing

PR #69, opened 2020-10-09
Fix the `cacheKey` type

### LINKED ISSUE #50: [TypeScript] cacheKey arguments array isn't automatically typed

owenallenaz:
Fixes #50

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -2,19 +2,26 @@
 import mimicFn = require('mimic-fn');
 import mapAgeCleaner = require('map-age-cleaner');
 
-const cacheStore = new WeakMap<(...arguments_: any[]) => any>();
+type AnyFunction = (...args: any) => any;
+
+const cacheStore = new WeakMap<AnyFunction>();
+
+interface CacheStorageContent<ValueType> {
+	data: ValueType;
+	maxAge: number;
+}
+
 interface CacheStorage<KeyType, ValueType> {
 	has: (key: KeyType) => boolean;
-	get: (key: KeyType) => ValueType | undefined;
-	set: (key: KeyType, value: ValueType) => void;
+	get: (key: KeyType) => CacheStorageContent<ValueType> | undefined;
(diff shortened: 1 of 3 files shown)


fregante:
This needs to be rebased on `master`:

```
git fetch origin master && git merge origin/master
// or
git fetch upstream master && git merge upstream/master
```

owenallenaz:
Tests are fixed. In the test.ts there are a whole mess of @ts-expect-error, almost all of those are unnecessary if the initial fn is initialized with `..._arguments: any` this way you're telling TS that you're gonna call that FN with all kinds of parameter configurations, which is what most of these tests are doing. I made this change only in the two fns that are failing as a result of my changes.

fregante (approved):
🎉 

@sindresorhus you might want to release this as v8 because it’s probably still not compatible with the prematurely-published 7.0.0-2

sindresorhus:
Thank you, @owenallenaz 🙌🏻

PR #72, opened 2021-03-19
Improve decorator example in readme

papb:
Since it's TS only, it is better to show the entire code example as it should be in TypeScript.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -206,8 +206,8 @@ Type: `object`
 
 Same as options for `mem()`.
 
-```js
-const mem = require('mem');
+```ts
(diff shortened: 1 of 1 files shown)


fregante:
Could you also extend the `mem.decorator` documentation? Like by adding a link to [the docs](https://www.typescriptlang.org/docs/handbook/decorators.html) and mentioning that Babel’s decorators are not compatible.

papb:
Hi @fregante, I don't have time to look into that now, sorry, feel free to take over the PR.

fregante:
Will do now

PR #73, opened 2021-03-19
Extend decorator documentation

fregante:
I just found out that the proposal only supports class methods :(

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -203,7 +203,7 @@ Returns a [decorator](https://github.com/tc39/proposal-decorators) to memoize cl
 Notes:
 
 - Only class methods and getters/setters can be memoized, not regular functions (they aren't part of the proposal);
-- Only [TypeScript’s decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#parameter-decorators) are supported, not Babel’s;
+- Only [TypeScript’s decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#parameter-decorators) are supported, not [Babel’s](https://babeljs.io/docs/en/babel-plugin-proposal-decorators), which use a different version of the [Decorators Proposal](https://github.com/tc39/proposal-decorators/issues/69);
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -198,7 +198,13 @@ Refer to the [caching strategies](#caching-strategy) section for more informatio
 
 ### mem.decorator(options)
 
-Returns a TypeScript decorator which memoizes the given function.
+Returns a [decorator](https://github.com/tc39/proposal-decorators) to memoize class methods.
+
+Notes:
+
+- Only class methods and getters/setters can be memoized, not regular functions (they aren't part of the proposal);
```
fregante, line 205: ?

```suggestion
- Only class methods and accessors can be memoized, not regular functions (they aren't part of the proposal);
```

fregante:
```suggestion
- Only class methods and accessors can be memoized, not regular functions ([they aren't part of the proposal]());
```

papb: I personally prefer `getters/setters`

readme.md
```diff
@@ -198,7 +198,13 @@ Refer to the [caching strategies](#caching-strategy) section for more informatio
...
+- Only [TypeScript’s decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#parameter-decorators) are supported, not Babel’s;
```
papb, line 206: Maybe you can link to Babel decorators as well?

readme.md
sindresorhus: `accessors` is the correct term, but I think `getters/setters` is better as it's more commonly known.

readme.md
fregante:
```suggestion
- Only [TypeScript’s decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#parameter-decorators) are supported, not [Babel’s](https://babeljs.io/docs/en/babel-plugin-proposal-decorators), which use a different version of the [Decorators Proposal](https://github.com/tc39/proposal-decorators/issues/69);
```

Issue #74, opened 2021-04-12
Values are memoised across class/prototype instances


notjosh:
Hi!

On the surface, it's kind of logical that (by default) values are instances are memoised across instances, because they're just function calls.

However: a) that seems almost never like intended behaviour, and b) there's [a test](https://github.com/sindresorhus/mem/blob/a42d7205cc7939e1738eda2a785fe1ec5bda2760/test.ts#L240-L256) that seems like it's explicitly trying to test for this.

Unfortunately, the test is only correct as a side effect. If you pass a new initial value into the constructor, you can break the logic (I've [fixed the test on my branch](https://github.com/notjosh/mem/commit/65bfb84346e834cfd77d59e89ca2025d479b5b84), but not sure how to fix the actual problem!)

Do you know how, either as a library consumer, or within the library itself we could get this working as expected?

fregante:
Thanks for trying the new decorator and opening an issue. Are you sure that test runs at all? `++this.index` sets it to undefined now I think

notjosh:
The tests run ok, but I think you're referring to not defining `index` on the class? It's using a shorthand syntax for assigning properties directly from the constructor, but is otherwise the same as before. (ref: https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/)

sindresorhus:
// @Richienb

Richienb:
Should mem be memoizing values return by a function across instances or for each instance?

sindresorhus:
For each instance by default at least. We could consider adding an option to opt into memorization across instances.

PR #75, opened 2021-04-19
Don't use the same memoized function across instances when `.decorator()` is used

Richienb:
Fixes: #74

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -68,15 +68,6 @@ interface Options<
 	readonly cache?: CacheStorage<CacheKeyType, ReturnType<FunctionToMemoize>>;
 }
 
-interface DecoratorOptions {
-	/**
-	Whether the same memoized function should be used in different instances or a different one should be initialised for each one.
-
-	@default false
-	*/
-	readonly isAcrossInstances?: boolean;
-}
-
 /**
 [Memoize](https://en.wikipedia.org/wiki/Memoization) functions - An optimization used to speed up consecutive function calls by caching the result of calls with identical input.
 
@@ -152,7 +143,7 @@ const mem = <
 export = mem;
 
 /**
-@returns A TypeScript decorator which memoizes the given function.
+@returns A [decorator](https://github.com/tc39/proposal-decorators) to memoize class methods or static class functions.
 
 @example
 ```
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -204,7 +204,14 @@ Returns a TypeScript decorator which memoizes the given function.
 
 Type: `object`
 
-Same as options for `mem()`.
+Same as options for `mem()` and the following:
+
+##### isAcrossInstances
```
fregante, line 209: When would this ever be useful? 🤷‍♂️ Instances are not meant to share a global states

fregante: The only situation that we're not covering anywhere are static methods (isAcrossInstances doesn't apply here either)

sindresorhus: @fregante It's not something you would need often, but sometimes you do calculations that don't actually depend on any state in the instance. I don't have an immediate need for this though, just thought someone might find it useful. Do you think it's worth keeping it?

fregante: > sometimes you do calculations that don't actually depend on any state in the instance

Isn't that the definition of a static method?

sindresorhus: Do decorators work on static methods? If so, we can indeed remove this option.

fregante: The docs don't include any static examples, but it doesn't mention "static" a few times: https://www.typescriptlang.org/docs/handbook/decorators.html#decorator-evaluation

Also it seems to pass:

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMGAnAzlAFASkQbwChFEIF04AbAUwDpK4BzLAIhQ2xwC5kBDaOKgCeiagDdelELyjUAJixwBuYolTUoIVElCRYCRFii9UjDT15ghAGkQAHVHDvVUUIQGlqQnplQwwjLZy1OgQfnZQgjwACo7OrkIAIiFhMBGCeEQkJGRgFDT0TKzsmLg8EJI0CsqqAL4qtYSEEJS86OiIAOLqGi4EqgACJZyqjD2c-dlqGlpILAAS1JQMLCokjSSDw7iqmDIwEIhj1Bq4k9nqmtqIC0sra4iNtUA

Screen Shot

sindresorhus: Alright let's remove the option and update the docs to mention support for static methods too, just to be user-friendly.

fregante: @Richienb I just merged the other PR that extended the docs for this. Update this branch before extending the docs further.

index.ts
```diff
@@ -141,7 +143,7 @@ const mem = <
 export = mem;
 
 /**
-@returns A TypeScript decorator which memoizes the given function.
+@returns A [decorator](https://github.com/tc39/proposal-decorators) to memoize class methods or static class functions.
```
sindresorhus, line 146:
```suggestion
@returns A [decorator](https://github.com/tc39/proposal-decorators) to memoize class methods or static class methods.
```

sindresorhus:
👍

PR #76, opened 2021-07-10
Require Node.js 12.20 and move to ESM

Richienb:


sindresorhus:
Tests are failing

.github/workflows/main.yml
```diff
@@ -10,9 +10,9 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
+          - 16
           - 14
           - 12
-          - 10
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
```
sindresorhus, line 18:
```suggestion
      - uses: actions/setup-node@v2
```

index.ts
```diff
@@ -96,24 +95,25 @@ memoized('bar');
...
+	}: Options<FunctionToMemoize, CacheKeyType> = {},
+): FunctionToMemoize {
 	if (typeof maxAge === 'number') {
 		// TODO: Drop after https://github.com/SamVerschueren/map-age-cleaner/issues/5
 		// @ts-expect-error
 		mapAgeCleaner(cache);
 	}
 
 	const memoized = function (this: any, ...arguments_) {
+		/* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-return */
```
sindresorhus, line 116: I think it's better to resolve the issue than to just ignore it.

index.ts
```diff
@@ -161,49 +156,51 @@ class Example {
...
-		throw new TypeError('The decorated value must be a function');
-	}
+	options: Options<FunctionToMemoize, CacheKeyType> = {},
+) {
+	return (
+		target: any,
+		propertyKey: string,
+		descriptor: PropertyDescriptor,
+	): void => {
+		const input = target[propertyKey]; // eslint-disable-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
```
Richienb, line 177: These errors must be ignored because `target: any` and there is currently no viable alternative.
https://stackoverflow.com/questions/68343539/typescript-definition-of-target-in-a-decorator

index.ts
```diff
@@ -96,63 +95,59 @@ memoized('bar');
...
 
-	const memoized = function (this: any, ...arguments_) {
-		const key = cacheKey ? cacheKey(arguments_) : arguments_[0];
+	const memoized = function (this: any, ...arguments_: Parameters<FunctionToMemoize>): ReturnType<FunctionToMemoize> {
+		const key = cacheKey ? cacheKey(arguments_) : arguments_[0] as CacheKeyType;
 
 		const cacheItem = cache.get(key);
 		if (cacheItem) {
-			return cacheItem.data;
+			return cacheItem.data; // eslint-disable-line @typescript-eslint/no-unsafe-return
```
Richienb, line 118: Because the return value of `AnyFunction` is `any` and `FunctionToMemoize extends AnyFunction`, the linter assumes that `ReturnType<FunctionToMemoize>` is `any`. However, the error is that `any` is **not** assignable to `ReturnType<FunctionToMemoize>`:

>   ✖  118:4  Unsafe return of type any from function with return type ReturnType<FunctionToMemoize>.  @typescript-eslint/no-unsafe-return

This might be a bug with typescript-eslint.

sindresorhus:
Looks good ;)

PR #78, opened 2021-07-15
Fix using multiple decorators on the same instance

### LINKED ISSUE #77: Memoized methods of the same instance are mixed
If two functions are decorated of the same instance, their memoized values get mixed with the latest version (`v9.0.0`).

Take for example:

```typescript
class TestClass {
	@memDecorator()
	fn1() {
		return 1;
	}

	@memDecorator()
	counter2() {
		return 2;
	}
}

const instance = new TestClass();
console.log(instance.fn1()); // prints 1
console.log(instance.fn2()); // also prints 1 !!!
```

I would expect the second call to return 2, but it returns 1.

The memoized function in my understanding should be connected 1:1 to the _combination_ of _instance_ and _propterty_, whereas it is currently connected 1:1 to _instances_ (#74).

I guess a little bit more complicated internal data structure is needed instead of the current implementation of `decoratorInstanceMap`, will open a PR that adds a test and demonstrates a possible solution.

bencehornak:
Fixes #77

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -3,8 +3,6 @@ import mapAgeCleaner from 'map-age-cleaner';
 
 type AnyFunction = (...arguments_: any) => any;
 
-const decoratorInstanceMap = new WeakMap();
-
 const cacheStore = new WeakMap<AnyFunction, CacheStorage<any, any>>();
 
 interface CacheStorageContent<ValueType> {
@@ -169,6 +167,8 @@ export function memDecorator<
 >(
 	options: Options<FunctionToMemoize, CacheKeyType> = {},
 ) {
+	const instanceMap = new WeakMap();
+
 	return (
 		target: any,
 		propertyKey: string,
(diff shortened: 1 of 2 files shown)


index.ts
```diff
@@ -3,7 +3,18 @@ import mapAgeCleaner from 'map-age-cleaner';
...
+_propertyDescriptor_ to a memoized function.
+
+The memoized function that belongs to a specific instance and a specific
+propertyDescriptor could theoretically be accessed as follows:
+
+```
+const mem = decoratorInstanceMap.get(instance)?.get(propertyDescriptor)
+```
+ */
+const decoratorInstanceMap = new WeakMap<object, WeakMap<PropertyDescriptor, unknown>>(); // eslint-disable-line @typescript-eslint/ban-types
```
fregante, line 17: I think the solution is to just move the weak map to the decorator function, but I'm not sure.

sindresorhus:
// @Richienb

index.ts
bencehornak: That's actually a pretty good idea! Tested it, works!

index.ts
```diff
@@ -169,6 +167,8 @@ export function memDecorator<
 >(
 	options: Options<FunctionToMemoize, CacheKeyType> = {},
 ) {
+	const decoratorInstanceMap = new WeakMap();
```
Richienb, line 170:
```suggestion
	const instanceMap = new WeakMap();
```

test.ts
```diff
@@ -239,19 +239,24 @@ test('prototype support', t => {
...
+			return returnValue2;
 		}
 	}
 
 	const alpha = new TestClass();
 	t.is(alpha.counter(), 1);
 	t.is(alpha.counter(), 1, 'The method should be memoized');
-
-	returnValue++;
+	t.is(alpha.counter2(), 101, 'Different methods should be independent');
```
Richienb, line 259:
```suggestion
	t.is(alpha.counter2(), 101, 'The method should be memoized seperately from the other one');
```

Richienb (changes requested):

Issue #81, opened 2021-07-19
Cache maxAge is evaluated wrong for async functions


krystofwoldrich:
I've noticed, that when memoizing long lasting async functions, the `maxAge` is evaluated relative to the moment when function call started. But in case of sync functions it is evaluated relative to the moment the function was resolved.

**Expected:** Sync functions as is. Async functions (functions that return promise) `maxAge` should be evaluated relative to when promise is resolved.

**Current:** Async functions `maxAge` is evaluated relative to the moment when Promise is returned. (Not resolved, as I would expect!)

**Fix:** Dirty fix is add `await`, but then all memoized function will be async. Clean fix is to change beaviour base on function return type. If Promise then wait till it's resolved. Else as is now.

I really appriciate your work, it's great library, just I can't use it because of this unexpected behaviour. Or if I miss understood something and it's working correctly I'm in advance sorry.

fregante:
`mem` cache the result of the function, which in this case is just a (Promise) object. What the object does isn't important as far as `mem` is concerned. What you're looking for is `p-memoize`, which wraps `mem`. You can probably open an issue there (or hope that Sindre can transfer this one to that repo first)

krystofwoldrich:
I understand that the function returns `Promise` so the behaviour is correct for sync code. But not for async, where unresolved Promise is just "not finished" function.

And since p-memoize is build on top of mem the same unexpected behaviour occours there too.

fregante:
> unresolved Promise is just "not finished" function.

No, Promises are objects that are returned immediately. `mem` stores what the function returns, which is an object. If you want Promise-aware memoization, `p-memoize` exists for that reason.

> And since p-memoize is build on top of mem the same unexpected behaviour occours there too.

I didn't say that it works there, but that it can be asked/reported there as it would be more appropriate

krystofwoldrich:
Okay. Then the mention in doc about `works fine with async fucntions` should dissapear.

I was asking about this, because in JS `async functions` and functions returning `Promise`, could be looked at as the same. But their intention is very different (in one promsie is the result, in one the resolved Promise is the wanted result), I think you would agree. But this library takes it as the same.

Issue #79, opened 2021-07-16
When first call encounter exception, the function always return exception


zacyang:
Notice the cache did not distinguish error from regular return, meaning if first try failed, all the subsequent call will fai.
Can we add a configure to allow only cache the success return?

fregante:
The result of functions isn't stored if they throw errors, but **rejected promises** are indeed stored. What you're asking is covered by `p-memoize`, which is a `mem` wrapper

https://github.com/sindresorhus/p-memoize#cachepromiserejection

Issue #49, opened 2019-11-21
mem vs. p-memoize


ghost:
mem supports promises, and the usage looks identical to p-memoize for the case of promises, which is listed in related packages. If my use-case is promise-only, would you recommend to switch to p-memoize, or should I keep using mem? 

In general it would be nice to clarify the relationship between the two packages in the README.

Thanks!

fregante:
Indeed currently they’re identical or nearly so. You can use `mem` for promises and viceversa. 

I recently opened a PR on `p-memoize` to add support for an asynchronous cache to differentiate the two. This would let you store the memoized values in a database, file system or any other asynchronous storage (via `keyv` package for example)

If you just need a regular `Map` storage you can continue to use `mem` with promises and async functions.

fregante:
@sindresorhus I wonder if it’s worth keeping two packages at all. The only difference will be the cache type (synchronous or asynchronous) and that can probably  detected at runtime by checking whether `.has()` returns a promise instead of a Boolean (or via option)

andreasvirkus:
Does `mem` also support ignoring promise rejections or is that unique to `p-memoize` (configurable via an option)? If they differ in that behaviour, then it seems like a thing worthy of highlighting.

sindresorhus:
@andreasvirkus It's not in the docs, so it does not.

PR #82, opened 2021-07-19
Mention `p-memoize` in the readme intro

fregante:
Related:

-  https://github.com/sindresorhus/mem/issues/49
-  https://github.com/sindresorhus/mem/issues/79
-  https://github.com/sindresorhus/mem/issues/81


krystofwoldrich:
I would also specified the behaviour with `async functions` since at the moment the `readme.md` says it work fine with them. :D

fregante:
That's the whole point of this new paragraph 😃

fregante:
Good point, I didn't remember we had that example

PR #91, opened 2023-05-04
Update dependencies

fregante:
https://github.com/SamVerschueren/map-age-cleaner/pull/7 / https://github.com/SamVerschueren/map-age-cleaner/releases/tag/v0.2.0 have been out for… ahem… a while. They should be updated.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -5,23 +5,23 @@ type AnyFunction = (...arguments_: any) => any;
 
 const cacheStore = new WeakMap<AnyFunction, CacheStorage<any, any>>();
 
-interface CacheStorageContent<ValueType> {
+type CacheStorageContent<ValueType> = {
 	data: ValueType;
 	maxAge: number;
-}
+};
 
-interface CacheStorage<KeyType, ValueType> {
+type CacheStorage<KeyType, ValueType> = {
 	has: (key: KeyType) => boolean;
 	get: (key: KeyType) => CacheStorageContent<ValueType> | undefined;
 	set: (key: KeyType, value: CacheStorageContent<ValueType>) => void;
 	delete: (key: KeyType) => void;
 	clear?: () => void;
-}
+};
 
-export interface Options<
+export type Options<
 	FunctionToMemoize extends AnyFunction,
 	CacheKeyType,
-> {
+> = {
 	/**
 	Milliseconds until the cache expires.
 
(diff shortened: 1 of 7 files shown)


index.ts
```diff
@@ -5,23 +5,23 @@ type AnyFunction = (...arguments_: any) => any;
 
 const cacheStore = new WeakMap<AnyFunction, CacheStorage<any, any>>();
 
-interface CacheStorageContent<ValueType> {
+type CacheStorageContent<ValueType> = {
```
fregante, line 8: Is this a breaking change?

package.json
```diff
@@ -38,30 +38,25 @@
...
-		"@types/serialize-javascript": "^4.0.0",
-		"ava": "^3.15.0",
-		"del-cli": "^3.0.1",
-		"delay": "^4.4.0",
-		"serialize-javascript": "^5.0.1",
-		"ts-node": "^10.1.0",
-		"tsd": "^0.13.1",
-		"typescript": "^4.3.5",
-		"xo": "^0.41.0"
+		"@sindresorhus/tsconfig": "^3.0.1",
```
fregante, line 45: This changes the output:

```
file:///home/runner/work/mem/mem/node_modules/xo/node_modules/meow/index.js:85
			flag.default = flag.default ?? [];
			                             ^

SyntaxError: Unexpected token '?'
```

Is it time to drop Node 12 while we're at it? Last breaking version was almost 2 years ago https://github.com/sindresorhus/mem/releases/tag/v9.0.0

package.json
```diff
@@ -38,30 +38,25 @@
...
+		"xo": "^0.54.2"
 	},
 	"ava": {
-		"timeout": "1m",
 		"extensions": {
 			"ts": "module"
 		},
-		"nonSemVerExperiments": {
-			"configurableModuleFormat": true
-		},
```
fregante, line 64: AVA Config copied from https://github.com/sindresorhus/p-queue

index.ts
sindresorhus: Should not be, no.

package.json
sindresorhus: Yes, we can target Node.js 16.

fregante:
I think the “tsx” package offers a better solution, but I think it doesn’t type-check. Do you want to try it? We’d have to add “test.ts” to tsconfig

sindresorhus:
I think this is the fix: https://github.com/sindresorhus/got/blob/main/tsconfig.json#L17-L21

tsconfig.json
```diff
@@ -6,5 +6,8 @@
 	},
 	"files": [
 		"index.ts"
-	]
+	],
+	"ts-node": {
+		"transpileOnly": true
```
fregante, line 11: This fixes the tests in <=18, but Node 20 tests still broken even with the full

```json
		"transpileOnly": true
		"files": true,
		"experimentalResolver": true
```

That's why I had initially excluded them

PR #94, opened 2023-11-14
Use custom cache expiration logic

### LINKED ISSUE #92: Replace/inline dead `map-age-cleaner` package
The package is outdated and it has an old [p-defer](https://github.com/sindresorhus/p-defer/releases) dependency:

- https://github.com/SamVerschueren/map-age-cleaner/issues/9
- https://github.com/SamVerschueren/map-age-cleaner/blob/0f9f8631510815dd7f567372461fa923c25f2c46/package.json#L38

Related:

- https://github.com/sindresorhus/mem/pull/91

sindresorhus:
Fixes #92

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,9 +1,9 @@
 import mimicFunction from 'mimic-function';
-import mapAgeCleaner from 'map-age-cleaner';
 
-type AnyFunction = (...arguments_: readonly any[]) => any;
+type AnyFunction = (...arguments_: readonly any[]) => unknown;
 
 const cacheStore = new WeakMap<AnyFunction, CacheStorage<any, any>>();
+const cacheTimerStore = new WeakMap<AnyFunction, Set<number>>();
 
 type CacheStorageContent<ValueType> = {
 	data: ValueType;
@@ -104,8 +104,8 @@ export default function mem<
 		maxAge,
 	}: Options<FunctionToMemoize, CacheKeyType> = {},
 ): FunctionToMemoize {
-	if (typeof maxAge === 'number') {
-		mapAgeCleaner(cache as unknown as Map<CacheKeyType, ReturnType<FunctionToMemoize>>);
+	if (typeof maxAge === 'number' && maxAge <= 0) {
+		return fn;
 	}
 
 	const memoized = function (this: any, ...arguments_: Parameters<FunctionToMemoize>): ReturnType<FunctionToMemoize> {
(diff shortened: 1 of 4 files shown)


index.ts
```diff
@@ -123,6 +123,18 @@ export default function mem<
...
 
+		if (typeof maxAge === 'number' && maxAge !== Number.POSITIVE_INFINITY) {
+			const timer = setTimeout(() => {
+				cache.delete(key);
+			}, maxAge);
+
+			timer.unref?.();
+
+			const timers = cacheTimerStore.get(fn) ?? new Set();
+			timers.add(timer as any); // eslint-disable-line @typescript-eslint/no-unsafe-argument
```
fregante, line 134: Are you looking for this? 📎 

```suggestion
			const timers = cacheTimerStore.get(fn) ?? new Set<NodeJS.Timeout>();
			timers.add(timer);
```

fregante (approved):

index.ts
```diff
@@ -104,8 +104,8 @@ export default function mem<
 		maxAge,
 	}: Options<FunctionToMemoize, CacheKeyType> = {},
 ): FunctionToMemoize {
-	if (typeof maxAge === 'number') {
-		mapAgeCleaner(cache as unknown as Map<CacheKeyType, ReturnType<FunctionToMemoize>>);
+	if (typeof maxAge === 'number' && maxAge <= 0) {
+		return fn;
```
fregante, line 108: This feels a little icky. <0 should probably be an error. Who memoizes a function with a calculated maxAge?

index.ts
sindresorhus: This package is environment independent, so I didn't want to use any Node.js specific types.

index.ts
sindresorhus: Agreed. Will do.

fregante: Similar situation in https://github.com/Richienb/batched-function#delay

- delay: undefined -> trigger in next micro task
- delay: 0 -> trigger the same way `setTimeout(fn, 0)` would

Either way not pretty

index.ts
fregante: I don’t think this type makes it into `distribution`?

index.ts
```diff
@@ -1,9 +1,9 @@
 import mimicFunction from 'mimic-function';
-import mapAgeCleaner from 'map-age-cleaner';
 
-type AnyFunction = (...arguments_: readonly any[]) => any;
+type AnyFunction = (...arguments_: readonly any[]) => unknown;
```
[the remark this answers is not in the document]
sindresorhus: It's going to be a major bump though. I think it's fine unless you know specific cases it will break.

index.ts
sindresorhus: Correct. It's technically fine, just my preference not to use it as it.

fregante: Can the same code work with Set<number>? The WeakMap type above uses it. This would let you move the assertion to `setTimeout() as number` instead of using any and eslint-disable

fregante:
```suggestion
			const timers = cacheTimerStore.get(fn) ?? new Set<number>();
			timers.add(timer as number);
```

PR #98, opened 2024-02-05
fix typescript typing reference in package.json

Goobles:
it seems that the types property was removed from the package.json here:
https://github.com/sindresorhus/memoize/commit/3afdfaf4f7862f835d01f250779091fe66e00f32#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L23

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
+	"types": "./dist/index.d.ts",
 	"exports": {
 		"types": "./dist/index.d.ts",
 		"default": "./dist/index.js"


fregante:
What is this for? Types already work without it

sindresorhus:
This is a problem with your tsconfig, not this package.

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm

Goobles:
@sindresorhus In that case I will not be able to import this package into an non-pure-ESM angular library project. I've switched over to the commonjs compatible `moize` instead.

Pure ESM is just not always an option, since all downstream dependencies then need to be patched as well, all imports refactored to include a '.js' extension, kind of sad how this causes a schism in the package ecosystem.

paul-vd:
> This is a problem with your tsconfig, not this package.
> 
> https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm

Why does npm not detect types? 
image

perry-mitchell:
Sure, but why avoid the defacto root types property? What would it cost to support it?

Not sure what moral ground is gained by dismissing it.

Issue #93, opened 2023-06-29
memDecorator calls getters immediately


neodon:
Hello,

I'm running into an issue memoizing a getter that relies on configuration set in the constructor. It appears that this line in `memDecorator` calls the getter immediately in addition to retrieving the getter function:

```typescript
const input = target[propertyKey]; // eslint-disable-line @typescript-eslint/no-unsafe-assignment
```

Example:

```typescript
import { memDecorator } from 'mem'

export class TestMemoizeGetter {
  constructorCalled?: boolean

  constructor() {
    this.constructorCalled = true
  }

  @memDecorator()
  public get someGetter() {
    if (!this.constructorCalled) {
      throw new Error('constructor has not been called')
    }

    return Math.random()
  }
}

const instance = new TestMemoizeGetter()

assert.doesNotThrow(() => {
  const value = instance.someGetter
  assert.strictEqual(instance.someGetter, value)
})
```

Output:

```
Error: constructor has not been called
    at Object.get someGetter (/home/neodon/p/mem-getters/index.ts:14:13)
    at file:///home/neodon/p/mem-getters/node_modules/mem/dist/index.js:84:29
    at __decorateClass (file:///home/neodon/p/mem-getters/index.ts:1:374)
    at <anonymous> (/home/neodon/p/mem-getters/index.ts:12:14)
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)
```

I messed around with the implementation for `memDecorator` and had some luck by using `descriptor.value` for functions and `descripter.get` for getter functions. It created some more problems though around `this` bindings. It seems that getter functions have a lot of voodoo baked in.

Let me know if you see value in this getting fixed. I can create a pull request with a failing test, and add some of the partial fix I figured out.

Also, the same issue exists with `p-memoize`, but it seemed better to start here since getters are synchronous.

fregante:
I think you’re right, that line does call the getter right when the decorator is first called:

https://github.com/sindresorhus/mem/blob/22c4b5e8fb033532eb070e24b26521e7898dfd77/index.ts#L177

`instance.prop` will immediately call the `get prop` function. 

mem does not support getters at this point, I suppose. We should use the information in the descriptor (`descriptor.value` and `descriptor.get`) instead of that line

fregante:
> the same issue exists with p-memoize

It’s best to also open an issue there and link it to this one so people from both packages can contribute solutions.

PR #104, opened 2024-12-04
Fix Issue #93: Prevent Getters from Being Called During Decoration and Preserve this Context

Smoke3785:
## Description:

Howdy everyone,

This PR addresses **issue #93**, where getters were being invoked during the decoration phase, leading to unintended side effects such as ReferenceError due to premature access before class initialization. Previously, attempts by @Neodon to resolve this issue were effective in some cases but failed to preserve correct `this` context.

## Proposed Solution:

To circumvent the loss of `this` context when reassigning getter implementation, I simply wrapped the original getter in a non-memoized function. This wrapped function is then assigned as the new getter, ensuring that the getter logic executes only when accessed without interfering with class instantiation or dependencies.

## Key Points:

1. Method Memoization:
 - Each instance method decorated with `@memoizeDecorator` now has its own memoized function stored using a unique Symbol so that memoization caches are per-instance.
2. Getter Handling:
 - Currently, getters are wrapped in a function that caches their returned value using a unique Symbol. However, this leads to stale data if the underlying properties change after the first access. This is how memoization is meant to work, but it isn't completely intuitive.
  - It might be worth adding a parameter to the options that accepts a callback that, as a side effect, clears the cache for the given method. I haven't tested passing in a reference to the method on the prototype, perhaps that works as well.
3. Weakmap Usage
  - I didn't try to replicate your weakmap strategy to store the cache. If you'd like to replace the use of symbol with a weakmap, by all means.
  
## Testing
The current implementation passes all existing tests - it may be worth adding some to ensure `this` is preserved properly. I'm not sure what specific issues @Neodon ran into preserving `this`, but the following code behaves as expected: 
```tsx
class Example {
  index = 0;

  @memoizeDecorator()
  counter(s: string) {
    return ++this.index;
  }

  @memoizeDecorator()
  get count() {
    return this.index + 1;
  }
}

const test = new Example();

console.log(test.count); // Expected: 1
test.index = 1;
console.log(test.count); // Expected: 2

console.log(test.counter("a")); // Expected: 2
console.log(test.counter("a")); // Expected: 2 (cached)
console.log(test.counter("b")); // Expected: 3
console.log(test.counter("c")); // Expected: 4
console.log(test.counter("a")); // Expected: 2 (cached)
```

Hope this helps!
Owen Rossi-Keen, [Iliad.dev](https://iliad.dev/)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -18,10 +18,7 @@ type CacheStorage<KeyType, ValueType> = {
 	clear?: () => void;
 };
 
-export type Options<
-	FunctionToMemoize extends AnyFunction,
-	CacheKeyType,
-> = {
+export type Options<FunctionToMemoize extends AnyFunction, CacheKeyType> = {
 	/**
 	Milliseconds until the cache entry expires.
 
@@ -54,7 +51,9 @@ export type Options<
 	@default arguments_ => arguments_[0]
 	@example arguments_ => JSON.stringify(arguments_)
 	*/
-	readonly cacheKey?: (arguments_: Parameters<FunctionToMemoize>) => CacheKeyType;
+	readonly cacheKey?: (
+		arguments_: Parameters<FunctionToMemoize>,
+	) => CacheKeyType;
 
 	/**
 	Use a different cache storage. Must implement the following methods: `.has(key)`, `.get(key)`, `.set(key, value)`, `.delete(key)`, and optionally `.clear()`. You could for example use a `WeakMap` instead or [`quick-lru`](https://github.com/sindresorhus/quick-lru) for a LRU cache.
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for working on this

index.ts
```diff
@@ -226,11 +255,13 @@ Clear all cached data of a memoized function.
 export function memoizeClear(function_: AnyFunction): void {
 	const cache = cacheStore.get(function_);
 	if (!cache) {
-		throw new TypeError('Can\'t clear a function that was not memoized!');
+		// eslint-disable-next-line @typescript-eslint/quotes
```
sindresorhus, line 258: The lint rule should not be ignored.

index.ts
```diff
@@ -18,10 +18,7 @@ type CacheStorage<KeyType, ValueType> = {
 	clear?: () => void;
 };
 
-export type Options<
-	FunctionToMemoize extends AnyFunction,
-	CacheKeyType,
-> = {
+export type Options<FunctionToMemoize extends AnyFunction, CacheKeyType> = {
```
sindresorhus, line 21: Don't make unrelated changes. That makes it hard to review.

index.ts
```diff
@@ -187,33 +198,51 @@ class ExampleWithOptions {
...
-		if (typeof input !== 'function') {
-			throw new TypeError('The decorated value must be a function');
+		if (typeof originalMethod !== 'function') {
+			throw new TypeError('The decorated value must be a function or a getter');
 		}
 
-		delete descriptor.value;
-		delete descriptor.writable;
+		if (!isGetter) {
+			const memoizedKey = Symbol(`__memoized_${String(propertyKey)}`);
```
sindresorhus, line 217: This won't work. Not all symbols have a text description.

sindresorhus (changes requested):

sindresorhus:
It should use WeakMap, not Symbol.

index.ts
Smoke3785: Got it, I think my linter got to that one

Smoke3785:
I'll implement with weakmap and resubmit. Thanks for taking a look!

index.ts
Smoke3785: What do you have your linter line width set to?

index.ts
Smoke3785: Got it, my formatter seems pretty opinionated - is there a linter / formatter configuration that I'm missing / not activating?

Smoke3785:
Hey Sindre. The cache should be stored on each instance in a WeakMap. It passes tests and seems to work as far as I'm able to test, but I am not confident my implementation plays well with the cache invalidation / key calculation options.

Do you have any suggestions for methodDecorator-specific test-cases to try?

I hope this addresses your concerns. Thank you again for reviewing, 
Owen Rossi-Keen, [Iliad.dev](https://iliad.dev/?from=open-source)

PR #115, opened 2026-07-21
Add `isMemoized()` and `Memoized` type

### LINKED ISSUE #114: Tagged return type
I want my helper to only accept memoized functions, but unlike `_.memoize`, `memoize` attempts to mimic the input function and type entirely.

An easy/ugly way would be to return `Object.assign(fn, {isMemoized: true})` but I'd be happy to just get a `Tagged<FunctionToMemoize, 'memoized'>` type of sorts

sindresorhus:
The return type of `memoize()` is unchanged, so the tag is only earned after the runtime guard.

Fixes #114

---

Why not an `isMemoized` property on the returned function? It changes the shape of the function, which defeats the point of mimicking the input exactly. It's also trivially forgeable and would show up in spreads, `Object.keys`, and snapshots. The information already exists internally in a `WeakMap`, so the check needs no extra surface.

Why not brand the return type of `memoize()` by default? It would leak `Memoized<…>` into every hover and error message, and can cause declaration-emit friction for consumers re-exporting a memoized function. A brand on the return value is also just an assertion, whereas the guard reflects reality: `memoize(fn, {maxAge: 0})` returns the input unchanged since nothing is cached, and `isMemoized()` correctly reports `false` for it.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -9,6 +9,28 @@ const cacheStore = new WeakMap<AnyFunction, CacheLike<unknown, unknown>>();
 const cacheTimerStore = new WeakMap<AnyFunction, Set<Timer>>();
 const cacheKeyStore = new WeakMap<AnyFunction, (arguments_: readonly any[]) => unknown>();
 
+declare const memoizedBrand: unique symbol;
+
+/**
+A function that has been verified as memoized by `isMemoized()`.
+
+Use it to accept only memoized functions. Since `memoize()` mimics the given function entirely, the only way to get this type is through the `isMemoized()` guard.
+
+@example
+```
+import memoize, {isMemoized, type Memoized} from 'memoize';
+
+const use = (function_: Memoized<(text: string) => boolean>) => function_('foo');
+
+const memoized = memoize((text: string) => Boolean(text));
+
+if (isMemoized(memoized)) {
+	use(memoized);
+}
+```
+*/
+export type Memoized<FunctionToMemoize extends AnyFunction> = FunctionToMemoize & {readonly [memoizedBrand]: true};
+
 export type CacheItem<ValueType> = {
 	data: ValueType;
 	maxAge: number;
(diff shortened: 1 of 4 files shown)


fregante:
This makes accepting functions a chore, either via `as`:

```js
const memoized = memoize(fn);
fnThatAcceptsMemoizedOnly(memoized as Memoized)
```

or via:

```js
const memoized = memoize(fn);
if (isMemoized(memoized)) {
  fnThatAcceptsMemoizedOnly(memoized)
}
```

when it could be:

```js
const memoized = memoize(fn);
fnThatAcceptsMemoizedOnly(memoized)
```


I don't find this useful. If I have to use assertions or always-true conditions, I might as well have a local type.

> memoize(fn, {maxAge: 0})

I'd call that bad code. At any rate I'd call `isMemoized(memoize(args)) === false` surprising behavior

sindresorhus:
I'm happy to consider other suggestions, but I have explained what won't work.

fregante:
> It would leak `Memoized<…>` into every hover and error message, and can cause declaration-emit friction for consumers re-exporting a memoized function

I don't see this as an issue, every type "leaks". This also sounds like `Tagged` shouldn't exist (or be used by libraries)

> A brand on the return value is also just an assertion

Yes! That's the definition of tagged types.

The only reason to reject this is that `Memoized<>` reflects a behavior rather than a shape... but then again `Tagged<string, 'User'>` doesn't change the shape either.

fregante:
https://github.com/sindresorhus/debounce-fn/blob/6b255563ac728133b4f81f97f095dc702670ce3f/index.d.ts#L69

sindresorhus:
> https://github.com/sindresorhus/debounce-fn/blob/6b255563ac728133b4f81f97f095dc702670ce3f/index.d.ts#L69

Legacy. Had I done it today, it would have been a free function that accepted a debounced function.

